PR #10, opened 2020-01-28
Stop tracking Gemfile.lock

jacobthemyth:
While `Gemfile.lock` was removed in 8de62b9295a3c6a43ed86b1112d7cf4ac3f4fa40, it got added back in 18fb257694752d0a969798bbbbfeace0c3a87a16 somehow, even though it was ignored. I think maybe it has something to do with rebasing because the dates on those commits are backwards from their commit order 🤷‍♂. Anyway, https://github.com/Kajabi/sage/pull/4 needs to be deleted because `kajabi-products` currently has a hard dependency on the older version of Rack, even though it has a vulnerability :cry:, so this just deletes `Gemfile.lock` again.


cmcfadzean (approved):
Thanks so much for catching this!

PR #17, opened 2020-02-04
Jcm example folders

cmcfadzean:


app/views/sage/examples/elements/_element.html.erb
```diff
@@ -9,10 +9,10 @@
   </div>
   <h6 class="example__label">Preview</h6>
   <div class="example__preview">
-    <%= render "example_elements/#{title}/preview" %>
+    <%= render "sage/examples/elements/#{title}/preview" %>
   </div>
   <h6 class="example__label">Code</h6>
   <div class="example__code">
-    <pre class="prettyprint"><code><xmp><%= render "example_elements/#{title}/preview" %></xmp></code></pre>
+    <pre class="prettyprint"><code><xmp><%= render "sage/examples/elements/#{title}/preview" %></xmp></code></pre>
```
teenwolfblitzer, line 16: Wondering if the `<xmp>` tag is necessary here, since we're already wrapping with `code` and `pre`.

teenwolfblitzer (approved):
😎 One non-blocking question about the use of the `xmp` tag, but looks good to me!

app/views/sage/examples/elements/_element.html.erb
cmcfadzean: I cent remember the exact reason but I believe that since its HTML code we have to add that or else we would need to swap out all the brackets and such to render the code correctly. Might be completely wrong on this but it breaks rendering when I try removing it so might be worth digging in on when we have some time

teenwolfblitzer: Ah, I see, maybe it has to do with the `xmp` monospace font rendering white space. Probably worth looking into down the road, since the tag is officially deprecated from HTML at this point.

teenwolfblitzer:
Follow-up issue added to investigate `xmp` tag removal:  #18

PR #20, opened 2020-02-06
where to run generator

cmcfadzean:
There was some confusion around where to run the generators inside the gem. This is a small change to update that in the docs and also moves the documentation instructions into the proper index placement.


teenwolfblitzer (approved):
👍 Looks good to me!

PR #48, opened 2020-02-19
Color palette updates

### LINKED ISSUE #45: Investigate use of HSL colors over hex values
`HSL` color values would allow us to more easily adjust saturation and lightness variations to differentiate interactive states. 

This would apply only to the Sage system color palette, and we should still provide hex/RGB equivalents in documentation as a fallback.

- https://css-tricks.com/hsl-hsla-is-great-for-programmatic-color-control/

teenwolfblitzer:
## Description
- Updates color palette design token with HSL values
- Modifies classnames and structure in color palette documentation
- Fixes missing `lang` attribute on HTML

## Related issues
- Closes #45


teenwolfblitzer:
Closing this in favor of keeping consistent hex code values

PR #52, opened 2020-02-25
Update Sage version

teenwolfblitzer:
## Description
Bumping version number to `0.4.1`.

Updated changelog: https://app.gitbook.com/@kajabi/s/sage/changelog


cmcfadzean (approved):
Make sure we update the changelog to reflect the changes if we have not done that already. Happy to help point you where to go if you need assistance with that.

PR #53, opened 2020-02-26
Live Markup, Style, JS Reloading

cmcfadzean:
Live Reloading


README.md
```diff
@@ -2,6 +2,29 @@
 
 The Sage Design System (SDS) is our single source of truth, providing everything you need to build great products for our customers. It is the culmination of designers and developers working together to give teams the ability to ship high-quality products faster.
 
+## Getting Started
+
+Start Guard for live reloading
+
+```console
+$ guard
```
teenwolfblitzer: Should this be `bundle exec guard` (or `guard --no-bundler-warning`)? When I run `guard` on its own, I get a warning to run one of those two commands to ensure dependencies are loaded.

teenwolfblitzer (approved):
👍 🔥 Looks good to me! One question to clarify about running `guard` below, seems to be running fine as-is, however.

philschanely:
@cmcfadzean and @monkeypox8 in working this more with Jay, the differences between your setups that work and mine and Andrew's that don't have been quite odd and perplexing. Due to the development experience nature of this task, (and perhaps thinking ahead to other features like `stylelint`) I'm proposing we switch and use `browser-sync` through a development-only NPM script. See the updated README.

Thoughts? 
cc: @pixelflips @AndrwM

cmcfadzean:
This works great.

cmcfadzean:
Seems great lets land this

package.json
```diff
@@ -0,0 +1,31 @@
...
+  "description": "The Sage Design System (SDS) is our single source of truth, providing everything you need to build great products for our customers. It is the culmination of designers and developers working together to give teams the ability to ship high-quality products faster.",
+  "main": "sage/pages/index",
+  "directories": {
+    "lib": "lib",
+    "test": "test"
+  },
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "dummy": "cd test/dummy && rails s",
+    "sync": "browser-sync start --port 4000 --proxy localhost:3000 --files 'app/**/*.*' --startPath 'sage/pages/index'",
```
teenwolfblitzer, line 13: I'm wondering if we should prevent browsersync from auto-opening a browser window by tacking on ` --no-open` as an option here, or maybe chaining in a `&& sleep <some number here>` on the `start` command. The rails server seems to take anywhere from 15-20 sec to spin up on my machine, so browsersync eagerly opens up the site, but we end up having to wait to reload the page. Only happens when starting, so it's definitely not a dealbreaker!

teenwolfblitzer (approved):
👍 🔥 This is fantastic! Have one question about if/when browsersync opens a new tab/window, but either way it's a great addition.

package.json
philschanely: Open to options here too. Let's ship this and see how we like it moving forward. I prefer for the tab to open right up, and would rather not guess at when to fire it up to sync with Rails; prefer to do that myself. If we feel differently over time lets fix with another PR.

PR #54, opened 2020-02-26
Update quick links module

philschanely:
Cleans up the Quick Links sidebar structure to use simpler BEM.


cmcfadzean (approved):
Like it

teenwolfblitzer (approved):
👍Looks good to me, great work!  One thing I'm wondering is whether we alter the link structure to use an unordered list rather than the current stack of anchor links. I'll leave it up to you whether we take this on in a follow-up or should even address at all.

philschanely:
@monkeypox8 definitely in favor of the ul standard. I'll work that in...

teenwolfblitzer (approved):
👍 Nice work!

PR #57, opened 2020-02-26
Jcm info in generators

### LINKED ISSUE #43: Modify generators to create descriptions with builds
In addition to the preview/code snippets and CSS that are generated using `rails generate sage_XXX`, we should also be able to add documentation/descriptions for each element/component/module in `lib/generators`.

cmcfadzean:
Adds a rules section to elements and components to tell developers about the gem. 
Also includes a section for do and dont info.
Adds this structure to the generators.
Screen Shot 2020-02-26 at 11 25 16 AM
Closes #43

### DIFF
--- a/app/views/sage/examples/components/_component.html.erb
+++ b/app/views/sage/examples/components/_component.html.erb
@@ -15,4 +15,6 @@
   <div class="example__code">
     <pre class="prettyprint"><code><%= CGI.unescapeHTML(CGI.escapeHTML render "sage/examples/components/#{title}/preview") %></code></pre>
   </div>
+  <h6 class="example__label">Rules</h6>
+  <%= render "sage/examples/components/#{title}/rules" %>
 </div>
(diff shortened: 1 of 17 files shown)


teenwolfblitzer (commented):
This is looking great! Should we also have the rules added to:
- Modules `lib/generators/sage_module/sage_module_generator.rb`
- Utilities: `lib/generators/sage_utility/sage_utility_generator.rb`

cmcfadzean:
@monkeypox8 how often do you feel that we will be putting rules around those two? My only thought was that Modules are not really reused so probably not a ton of uses to address and utilities are so versatile that the rules could vary so much might not be as helpful. Thoughts?

teenwolfblitzer:
> @monkeypox8 how often do you feel that we will be putting rules around those two? My only thought was that Modules are not really reused so probably not a ton of uses to address and utilities are so versatile that the rules could vary so much might not be as helpful. Thoughts?

Hmm, that's probably true. I was anticipating that sometime in the near future, modules may end up being adopted for unintended use cases. So the Dos/Don'ts could help guide around that, however unlikely it'd be. But after yesterday's discussions with the team, that may be a moot point since we'll be more loosely grouping "modules" and "components" together.

cmcfadzean:
Closed for better option

PR #61, opened 2020-02-28
Change To Objects

cmcfadzean:
Changes the wording used in the system from components and modules to simply having elements and objects. 
Updates the generator to reflect this
rails g sage_object OBJECT_NAME


teenwolfblitzer (approved):
👍 Looks good to me!

PR #62, opened 2020-02-29
layout to layout folder

cmcfadzean:
With the switch away from modules some of the new objects are better served in the layouts folder.


teenwolfblitzer (approved):
👍 Looks good to me!

PR #63, opened 2020-02-29
new status page

cmcfadzean:
status page similar to this: 
https://thumbprint.design/components/overview/
Screen Shot 2020-03-03 at 8 57 19 PM

### DIFF
--- a/app/assets/stylesheets/sage_docs/_status_key.scss
+++ b/app/assets/stylesheets/sage_docs/_status_key.scss
@@ -15,19 +15,21 @@
     height: .8rem;
     border-radius: 50%;
     &--done {
-
+      border-color: sage-color(sage);
+      background: sage-color(sage);
     }
     &--doing {
-
+      border-color: sage-color(orange);
     }
     &--todo {
-
+      border-color: sage-color(primary);
     }
     &--no {
-
+      background: sage-color(grey);
     }
     &--stop {
-
+      border-color: sage-color(red);
+      background: sage-color(red);
     }
   }
   &__text {
--- a/app/assets/stylesheets/sage_docs/_status_key.scss
+++ b/app/assets/stylesheets/sage_docs/_status_key.scss
@@ -1,5 +1,5 @@
 .status-key {
-  margin: sage-spacing(xs) sage-spacing();
+  margin: 0 sage-spacing();
   display: flex;
   align-items: center;
   &:first-of-type {
@@ -7,13 +7,14 @@
   }
   &__wrapper {
     display: flex;
+    margin-bottom: sage-spacing();
   }
   &__icon {
-    margin-right: sage-spacing(xs);
     border: 2px solid sage-color(grey);
     width: .8rem;
     height: .8rem;
     border-radius: 50%;
+    display: inline-block;
     &--done {
       border-color: sage-color(sage);
       background: sage-color(sage);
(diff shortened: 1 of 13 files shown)


app/assets/stylesheets/sage_docs/_status_table.scss
```diff
@@ -0,0 +1,42 @@
...
+      padding-right: 0;
+    }   
+  }
+  td {
+    padding: sage-spacing(xs) sage-spacing(sm);
+    vertical-align: top;
+    &:nth-child(1) {
+      font-weight: sage-font-weight(semibold);
+      padding-left: 0;
+      width: 10000px;
```
teenwolfblitzer, line 20: I'd like to see us move away from the pattern of using large fixed widths for spacing in the codebase. If necessary, we could instead set the `tr` to be a flex container and adjust the `flex-basis` accordingly.

app/assets/stylesheets/sage_docs/_status_table.scss
```diff
@@ -0,0 +1,42 @@
...
+    }
+    &:nth-child(2) {
+      border-bottom: sage-border();
+      width: 4000px;
```
teenwolfblitzer, line 24: Same here as above with the fixed width value.

app/views/sage/pages/status.html.erb
```diff
@@ -0,0 +1,57 @@
...
+        <li class="quick-links__item">
+          <a href="#<%= object[:title] %>" class="quick-links__link"><%= object[:title].titleize %></a>
+        </li>
+      <% end %>
+    </ul>
+  </div>
+<% end %>
+<div class="sage-panel">
+  <h3>Elements</h3>
+  <table class="status-table">
```
teenwolfblitzer, line 30: We can address this in a follow-up (Chris and I have some ideas for tweaks and making the table into a component), but the mobile view is breaking outside of the panel:

status-table-mobile

teenwolfblitzer (approved):
👍 This is looking great! I've added a few things to consider that can all be addressed in a follow-up.

app/assets/stylesheets/sage_docs/_status_table.scss
cmcfadzean: Love it. I think this gets us moving forward but lets put in issue into the backlog to revisit this.

app/views/sage/application/_status_key.html.erb
```diff
@@ -0,0 +1,22 @@
+<div class="status-key__wrapper">
```
philschanely, line 1: For cleaner BEM I'd suggest this change to `status-key-wrapper`.

app/assets/stylesheets/sage_docs/_status_key.scss
```diff
@@ -0,0 +1,42 @@
+.status-key {
+  margin: 0 sage-spacing();
+  display: flex;
+  align-items: center;
+  &:first-of-type {
+    margin-left: 0;
+  }
+  &__wrapper {
```
philschanely, line 8: Per comment below, this may need would also be adjusted to keep in sync.

philschanely (commented):
A few suggestions for cleaner BEM.

app/views/sage/application/_status_key.html.erb
cmcfadzean: Would you not consider it to be an element of the status-key block? Interested to hear more.

PR #64, opened 2020-02-29
Jcm sidebar items

cmcfadzean:
Rendering the elements and components in the sidebar.
Only showing sub links when in a section where they apply.


teenwolfblitzer (approved):
👍 🔥 This is looking great! I'm thinking in a follow-up we might want to take a look at making the sidebar links into nested unordered lists… not sure how easy it is to add in, but it should make the eventual accordion view much easier to implement once we get these other items out of the way.

PR #66, opened 2020-03-02
Loading indicators

### LINKED ISSUE #65: Loader & spinner elements
## Description
Create animated small & large loader/spinner elements to match designs.

Inspiration for "large": https://lottiefiles.com/web-player?lottie_url=https://assets9.lottiefiles.com/datafiles/aba45c7b75d547282b2dbdc97969412b/progress_bar.json
- WIP example: https://codepen.io/monkeypox8/pen/RwPVxLo?editors=1100

Screen Shot 2020-02-29 at 3 37 44 PM

teenwolfblitzer:
## Description
Adds loading bar and loading spinner indicators
- Can be attached to any container
- Visibility & animation is triggered by data attribute
- SVG images added for reference of url-encoded background images

### Steps for testing
1. Run `rails s` to view the Sage documentation
2. Navigate to "Elements" in the main sidebar
3. Scroll to "Loader"

## Related issues
- Fixes #65


cmcfadzean (approved):
This is great. 
I do like the local variables that reference the global variables. Seems very clean for working on it later.

app/assets/stylesheets/sage/patterns/elements/_loader.scss
```diff
@@ -0,0 +1,66 @@
+// ============================================
```
philschanely, line 1: Aside: Is this commenting style a particular convention documented somewhere? Curious about this style versus what's on line 41.

app/assets/stylesheets/sage/patterns/elements/_loader.scss
```diff
@@ -0,0 +1,66 @@
...
+// Variables
+// ============================================
+$sage-loading-bar-bg-color: sage-color(grey);
+$sage-loading-bar-height: 0.625rem; // 10px
```
philschanely, line 5: Aside: I would love to add a function that takes an integer such as `10` and converts it to `rem`. I have used this in some other projects so that I can use human-friendly values and have the machine convert them to flexible rem. Open to it?

app/assets/stylesheets/sage/patterns/elements/_loader.scss
```diff
@@ -0,0 +1,66 @@
...
+  }
+}
+
+
+/*------------------------------------*/
+/*-- loading indicators */
+/*------------------------------------*/
+
+.sage-loading--bar {
+  @include sage-loading();
```
philschanely: This is different from the approach to BEM I'm accustomed to. I'd expect we separate and stack when we use modifiers. Thoughts? See https://css-tricks.com/bem-101/

```
<div class="sage-loading sage-loading--bar"></div>
```

philschanely (approved):
A few practice questions but nothing blocking given conventions in place already.

app/assets/stylesheets/sage/patterns/elements/_loader.scss
teenwolfblitzer: Hm, good point. The style on line 41 is auto-generated by Sage, where this one I was hoping to eventually standardize as a multi-line comment. It's all part of a gist that I'm working on for a VSCode snippet file here: https://gist.github.com/monkeypox8/f6157c22820a7cb5042bf43e8d536807

app/assets/stylesheets/sage/patterns/elements/_loader.scss
teenwolfblitzer: Sounds good to me!

app/assets/stylesheets/sage/patterns/elements/_loader.scss
cmcfadzean: Interested to hear opinions on this one. I tend to prefer letting scss take care of things and limit the class stacking and simplify the markup but would love to hear thoughts on this in general. @pixelflips ? @AndrwM

teenwolfblitzer: Oof. Yeah, I should've refactored that when I started working on the spinner. Will make the change.

teenwolfblitzer: > Interested to hear opinions on this one. I tend to prefer letting scss take care of things and limit the class stacking and simplify the markup but would love to hear thoughts on this in general. @pixelflips ? @AndrwM

I'm all for the chained class approach when it comes to variants, especially if we discontinue use of `@extend`,  as the majority had been leaning in the survey. While it means adding more classes in HTML, it also means less "bloat" in variants, since they'll no longer need to duplicate parent styles. But I also don't think this means variants should be too granular, to the point where we're adding minute changes with multiple variants.

PR #69, opened 2020-03-03
Jcm docs

cmcfadzean:
Updating gem to remove the pages and moving the docs into the github wiki


teenwolfblitzer (approved):
👍 Looks good to me. This might be overkill, but I'm wondering if we should also add a link to the Sage sidebar nav for the wiki. The changelog link already takes you there, but it might not be obvious enough that the other documentation lives there.

PR #73, opened 2020-03-04
Grid cleanup

### LINKED ISSUE #55: Add `.sage-col` as grid column fallback
## Description
`.sage-col` was removed as an option when the grid was refactored. We should add this class back as a default that allows content to scale accordingly.

### Example
With Sage running, visit the following page and inspect the `.sage-panel` container on one of the palette groups. You may have to look at several.
- http://localhost:3000/sage/pages/color

teenwolfblitzer:
## Description
- Adjusts `flex-grow` property to auto-size `.sage-col` default/fallback column class
- Refactor and cleanup of styles

### Screenshots
|  Before   |  After  |
|--------|--------|
|grid-before|grid-after| 

## Related issues
- Resolves #55

### DIFF
--- a/app/assets/stylesheets/sage/layout/_grid.scss
+++ b/app/assets/stylesheets/sage/layout/_grid.scss
@@ -2,13 +2,11 @@
 // Variables
 // ============================================
 
-$sage-grid-columns-sm: 12;
-$sage-grid-columns-md: 12;
-$sage-grid-columns-lg: 12;
+$sage-grid-columns: 12 !default;
 
-$sage-grid-gap-sm: sage-spacing(sm);
-$sage-grid-gap-md: sage-spacing(sm);
-$sage-grid-gap-lg: sage-spacing(sm);
+$sage-grid-gap-sm: sage-spacing(sm) !default;
+$sage-grid-gap-md: sage-spacing(sm) !default;
+$sage-grid-gap-lg: sage-spacing(sm) !default;
 
 // ============================================
 // Rows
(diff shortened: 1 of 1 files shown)


cmcfadzean (approved):
Cool

app/assets/stylesheets/sage/layout/_grid.scss
```diff
@@ -27,59 +25,48 @@ $sage-grid-gap-lg: sage-spacing(sm);
...
-  }
-
-  @for $sage-i from 1 through $sage-grid-columns-md {
-    &--md-#{$sage-i} {
-      @extend .sage-col;
-    }
-  }
-
-  @for $sage-i from 1 through $sage-grid-columns-lg {
+  @for $sage-i from 1 through $sage-grid-columns {
```
philschanely, line 32: 😍 Nice refactor!

app/assets/stylesheets/sage/layout/_grid.scss
```diff
@@ -27,59 +25,48 @@ $sage-grid-gap-lg: sage-spacing(sm);
...
+    &--sm-#{$sage-i},
+    &--md-#{$sage-i},
     &--lg-#{$sage-i} {
       @extend .sage-col;
```
philschanely, line 36: Would love to see us move away from this extends approach... 😅

philschanely (approved):
Nice work! 🙌 

Aside: Interesting that we use 12 columns all the way. Perhaps we should rethink this eventually.

app/assets/stylesheets/sage/layout/_grid.scss
teenwolfblitzer: Agreed, I think once we're in a good spot with linting, I intend to sweep through and clean up some of the "old" CSS.

teenwolfblitzer:
> Nice work! 🙌
> 
> Aside: Interesting that we use 12 columns all the way. Perhaps we should rethink this eventually.

@philschanely Good point. Previously, I'd added the option for each breakpoint grouping to modify its individual grid column count, but there isn't a use case for it *just yet*. We'll have to keep an eye on it.

PR #72, opened 2020-03-06
Refactor side nav menu

teenwolfblitzer:
## Description
- [x] Fix BEM leak in `_sidebar` for the menu button
- [x] (optional) Add overlay/underlay and behavior to close menu on content click


app/assets/stylesheets/sage/patterns/objects/_sidebar.scss
```diff
@@ -25,156 +25,37 @@
...
     transition: transform 0.2s ease-out, opacity 0.15s ease-in-out;
+    opacity: 0;
 
     .sidebar-is-open & {
+      transform: none;
       box-shadow: sage-shadow();
       opacity: 1;
+    }
+
+    .example__preview & {
```
teenwolfblitzer, line 39: Prevents sample sidebar from being triggered by nav button. Open to suggestions on a more elegant solution.

app/assets/stylesheets/sage/patterns/elements/_menu_button.scss
```diff
@@ -0,0 +1,125 @@
...
+}
+
+.sage-menubutton__bar {
+  display: block;
+  height: $sage-menubtn-toggle-line-height;
+  background: currentColor;
+  border-radius: $sage-menubtn-toggle-radius;
+  transition: transform 0.2s ease-out, opacity 0.15s ease-in-out;
+
+  + .sage-menubutton__bar {
```
philschanely: Stylistically, is this any easier to read than `&:not(:first-child)` or do we gain any benefits from one versus the other? 🤔

app/assets/stylesheets/sage/patterns/objects/_sidebar.scss
philschanely: 🎉

app/views/sage/application/_assistant.html.erb
```diff
@@ -1,6 +1,6 @@
 <div class="sage-assistant">
-  <button class="sage-sidebar__toggle" aria-expanded="false" aria-controls="sage-sidebar-menu">
-    <span class="btn-bar"></span><span class="btn-bar"></span><span class="btn-bar"></span>
+  <button class="sage-menubutton" aria-expanded="false" aria-controls="sage-sidebar-menu" data-js-action="sage-sidebar__toggle">
```
philschanely: Seeing this now I wonder if the `data-js-action` needs the BEM structure. Is this used for style in any way? If not, I wonder if we refactor to `sage-sidebar-toggle` or `sage-sidebar-is-toggled`?

philschanely (approved):
Non-blocking comments

app/assets/stylesheets/sage/layout/_header.scss
```diff
@@ -0,0 +1,9 @@
+/* ==================================================
+  ** _header.scss
+
+
+================================================== */
+
+.sage-header {
+  z-index: sage-z_index(nav);
+}
```
AndrwM: Missing end of file newline. This is probably the most nittiest nit you could ever nit. – I don't see it as a big deal but I know that it can create some hiccups when programmatically processing files with `grep` or `cat` because lines not proceeded by a newline aren't considered valid lines.

Here's some background info on end of file newlines:
- https://stackoverflow.com/a/729795
- https://unix.stackexchange.com/a/18789

This can be fixed permanently by setting: `files.insertFinalNewline` as `true` your VS Code setup.

Will this ever actually cause us issues? I honestly don't know, maybe never.

app/views/sage/application/_assistant.html.erb
AndrwM: I think it would make more sense to use `data-js="name"` rather than `data-js-action="..."` these won't always relate to an action that the user will perform but it will always be indicative of a JS binding.

For example JS listening to a `textarea[data-js="comment-box"]` and then performing an action or JS changing the state of a element bound with `data-js="name"` based on an action the user has taken elsewhere on the page.

AndrwM (approved):
The JS binding could use to be refactored to `data-js` over `data-js-action`.

I don't think the newline issue is a big deal, just throwing a flag there.

app/assets/stylesheets/sage/layout/_header.scss
teenwolfblitzer: Good callout! I'd left this rule alone in the default lint config not fully understanding why it was necessary.

app/views/sage/application/_assistant.html.erb
teenwolfblitzer: >Seeing this now I wonder if the data-js-action needs the BEM structure. Is this used for style in any way? If not, I wonder if we refactor to sage-sidebar-toggle or sage-sidebar-is-toggled?

The BEM structure isn't needed, I'll make that change. Originally I was toying with the idea that `data-js-action` could be used to select the ID (or another data attribute) of the targeted element rather than via a body class as I'm doing now. All this in order to avoid the issue addressed [here](https://github.com/Kajabi/sage/pull/72/files#r389204420) where the example sidebar is also toggled by the menu button thanks to that body class. I'm wondering to what extent we should be writing and testing JS for each object/module as if they're to be used in multiple places… though presumably the sidebar should only be included once, so it's a little more ambiguous.

> I think it would make more sense to use data-js="name" rather than data-js-action="..." these won't always relate to an action that the user will perform but it will always be indicative of a JS binding.

Sounds good to me! I'll refactor.

app/assets/stylesheets/sage/patterns/elements/_menu_button.scss
teenwolfblitzer: Hmm, great question. For me personally, readability with a combinator feels less complicated than with `&:not(:first-child)`, but I'm definitely open to changing it up depending on what the group thinks.

test/dummy/app/assets/javascripts/docs.js
```diff
@@ -1,29 +1,74 @@
...
+  sageNavOverlay.addEventListener('click', function(e) {
+    resetSideNav();
   });
 
-  // sidebar key events
+
+  // Close overlay and menu on esc keypress
   document.addEventListener('keyup', function(e) {
-    var keyNum = 'which' in e ? e.which : e.keyCode;
+    const keyNum = 'which' in e ? e.which : e.keyCode;
```
pixelflips, line 66: @monkeypox8 - not sure if this applies right now, but I previously had issues in the kajabi-products repo in regards to using ES6. The build would fail when precompiling and it took me a bit to figure out what was going on (using let was the issue). Maybe more of an FYI, but it might be something we want to check with @cmcfadzean about since eventually, the Kajabi app will consume Sage. Not a show stopper at all, just thought I'd share.

teenwolfblitzer: Good call. I did try to deliberately limit my use of ES6 to `const` (since that's pretty much the only thing IE11 supports) until we have a chance to get Babel into the build process, so this _should_ work until that point 🤞.

pixelflips (approved):
👍 Nice work!

PR #80, opened 2020-03-09
Update quicklinks subheading styles and semantics

### LINKED ISSUE #78: Address semantics in quick-links (Status page)
Quick links markup with subheadings (such as on the Status page) should be considered from both an aesthetics standpoint (how do we want subheadings over groups of links to look?) and semantics standpoint (we should not use `li` as headings but instead end the list and use a subheading over a new list)

philschanely:
Closes #78 

| Before | After |
|---|---|
| Screen Shot 2020-03-06 at 7 43 17 PM | Screen Shot 2020-03-09 at 11 34 24 AM |


cmcfadzean (approved):
Dig it

PR #46, opened 2020-03-10
SCSS variable structure - integrate CSS custom properties part 1

### LINKED ISSUE #70: Combine mixins for global use
## Description
We're currently using mixins within the Sass modules where they're being consumed. Whenever possible, we should consolidate applicable mixins into the global `_mixins.scss` file for easier reference and reuse.

teenwolfblitzer:
## Description
Splitting this task into separate parts for easier consumption.

- Refactor and streamline use of Sass variables in stylesheets.
- Relocate mixins to global stylesheet

Eventually, we may implement additional themes for Sage, including dark mode. Replacing/augmenting with CSS custom props as necessary would be extremely helpful, though this is likely to be dependent on future IE11 support.

### Steps for testing
TBD

## Related Issues
- Resolves #70


app/assets/stylesheets/sage/core/_variables.scss
```diff
@@ -0,0 +1,118 @@
...
+$sage-form-element-spacing: 1rem !default;
+
+// ==================================================
+// BUTTONS
+// ==================================================
+
+// Base styles
+$sage-btn-transition: 0.1s ease-in-out !default;
+$sage-btn-border-width: 0.125rem !default;
+$sage-btn-shadow-base: 0 1px 2px rgba(0, 0, 0, 0.18), 0 1px 3px rgba(0, 0, 0, 0.1) !default;
```
teenwolfblitzer: I considered placing this into the `sage-shadow` design token but it felt like more of a one-off

app/assets/stylesheets/sage/core/_variables.scss
```diff
@@ -0,0 +1,118 @@
+:root {
+  --something: something;
```
teenwolfblitzer, line 2: Nothing lives here yet. Will revisit in part 2 and beyond.

app/assets/stylesheets/sage/core/_variables.scss
cmcfadzean: Might be worth going back to the designer and ensuring that we dont want to utilize one of our current shadows. This feels like a bit of a departure from the system and we should as much as possible avoid 1 offs.

app/assets/stylesheets/sage/core/_variables.scss
```diff
@@ -0,0 +1,118 @@
...
+// ==================================================
+
+// Colors
+$sage-checkbox-color-default: sage-color(grey, 400) !default;
+$sage-checkbox-color-checked: sage-color(primary) !default;
+$sage-checkbox-color-disabled: sage-color(grey, 200) !default;
+$sage-checkbox-color-disabled-checked: sage-color(primary, 200) !default;
+
+$sage-checkbox-size: 1.5rem !default;
+$sage-checkbox-border-radius: 0.25rem !default;
```
cmcfadzean: Might be worth seeing if this needs to be different then our default radius

app/assets/stylesheets/sage/patterns/elements/_switch.scss
```diff
@@ -1,25 +1,10 @@
...
-$sage-switch-toggle-size: 1.25rem;
-
-$sage-switch-focus-outline-spacing: sage-spacing(2xs);
-$sage-switch-focus-outline-width: 1;
-$sage-switch-focus-outline-color: currentColor;
-
-/*------------------------------------*/
-/*-- switch element */
-/*------------------------------------*/
+/* ==================================================
```
cmcfadzean, line 1: It might be worth putting a follow up to update the generators to include this type of comment.

cmcfadzean (approved):
Couple things to clarify

philschanely (approved):
🎉

app/assets/stylesheets/sage/core/_variables.scss
teenwolfblitzer: Good catch! This should actually be set using the default.

app/assets/stylesheets/sage/core/_variables.scss
teenwolfblitzer: Turns out the shadow values were updated in figma but the design tokens were using the old values. I'll go through and make sure they match.

app/assets/stylesheets/sage/patterns/elements/_switch.scss
teenwolfblitzer: Added #86 as an issue

PR #23, opened 2020-03-10
Add visual indicators for page navigation

pixelflips:
Indicators for active navigation are not displayed, making it difficult to visualize the current page state at a glance.

page-nav

Suggested solutions to explore:

- Render breadcrumb nav below the title
- Highlight current/active page in sidebar navigation

~~Adding active page states in the sidebar will most likely require remapping the existing page routing. For example, the current route `/pages/conventions` would go to `/guidelines/conventions`.~~

An 'overview' page may also need to be added when a user selects "Conventions" in the navigation, rather than redirecting to the first subitem. This would be especially necessary for a breadcrumb nav.

### DIFF
--- a/app/assets/stylesheets/sage/patterns/objects/_nav.scss
+++ b/app/assets/stylesheets/sage/patterns/objects/_nav.scss
@@ -2,9 +2,10 @@
   &__link {
     display: flex;
     align-items: center;
-    margin: sage-spacing(xs) 0;
+    margin: sage-spacing(2xs) 0;
     font-weight: 500;
     color: sage-color(charcoal, 400);
+    padding: sage-spacing(2xs) sage-spacing(xs);
     &:hover {
       color: sage-color(charcoal, 500);
       i {
@@ -19,12 +20,13 @@
         color: sage-color(primary);
       }
     }
-    &--active{
-      color: sage-color(primary);
+    &--active {
+      background: sage-color(grey, 300);
+      border-radius: sage-border(radius);
     }
     i {
-      font-size: sage-spacing();
       margin-right: sage-spacing(xs);
+      font-size: sage-spacing();
       color: sage-color(grey, 500);
     }
   }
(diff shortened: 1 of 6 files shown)


pixelflips:
@monkeypox8 @cmcfadzean - I'd like to get some early feedback on this approach before moving to review. It seems to work great, but any thoughts would be appreciated. Also note, the active state styles are just a first-round idea since at some point we'll be updating the design. Thanks in advance!

teenwolfblitzer (commented):
I'm liking where this is heading. I do think we'll need some sort of focus state in addition to the "active" style when navigating via keyboard. So far, the global pattern for controls adds a rounded outline similar to the buttons/checkboxes, but I'm not sure how that'd fit here. If you haven't already, I'd ask Chris or Sam about their thoughts.

Also, I'm getting a little ahead of things, but have you considered swapping padding for margin on `.sage-nav__link--sub` so that the click target and visual indicator is larger?

pixelflips:
@monkeypox8 - I definitely like the idea of expanding the click target of the `.sage-nav__link--sub`, I'll update that. Thanks!

I was initially thinking this PR would focus mostly on the "active" class being applied to the proper link in the sidebar. If we're going to be updating the design and/or restructuring in the future, do you think it's worth waiting so that we will only have to update/adjust a small number of styles when work begins on the new sidebar?

cc: @cmcfadzean

teenwolfblitzer:
> I was initially thinking this PR would focus mostly on the "active" class being applied to the proper link in the sidebar. If we're going to be updating the design and/or restructuring in the future, do you think it's worth waiting so that we will only have to update/adjust a small number of styles when work begins on the new sidebar?

Hmm, that's true, though it could be some time before the updated design is fully implemented. I think whatever tweaks you can make should be enough to hold us over until that point.

pixelflips:
> Hmm, that's true, though it could be some time before the updated design is fully implemented. I think whatever tweaks you can make should be enough to hold us over until that point.

@monkeypox8 - ok I think with that said, I will keep this PR small and related to the "active" class for now and then revisit once we start on the full design updates. Thanks for all of your input and suggestions!

philschanely (approved):
No concerns here. 👍 🎉

pixelflips:
@monkeypox8 - these changes have been merged into master. I had a Git issue, where it looks like my branch diverged. Not too sure what happened, but the exact changes are now in master.

I will close this PR and drop it into the landed column since the changes are actually merged.

cc: @philschanely @cmcfadzean @AndrwM

pixelflips:
Updates are merged in d808928 - closing this PR.

PR #89, opened 2020-03-12
Add sample avatar images

### LINKED ISSUE #88: Add Live Profile Card
## Task

Add new object to support [Live (Issue #13474)](https://github.com/Kajabi/kajabi-products/issues/13474)

[Design](https://www.figma.com/file/9bVs5GcQLQWrhjVYyGph68/%E2%9A%99%EF%B8%8F-Kajabi-Live-MASTER?node-id=2063%3A9167)

Screen Shot 2020-03-12 at 1 47 08 PM

## Proposed Architecture

- `sage-live-avatar` -- The circular image shown here and used elsewhere. Can scale to any size 128px and smaller. Variations include adding the colored ring around the image such as for the host and participant. See #93 
- `sage-live-user-control` -- Core component for user control buttons to include normal, hover, and inactive colors applied to icons. Variations include those for different button commands: video on/off, microphone on/off, hand, and options. See #91 
- `sage-live-profile-card` -- Card component that contains the avatar, name information, and user control bar. Variations include constructs for the hover and role-based states.

philschanely:
Related to #88 

- [x] Add `sage-live-avatar`

Link in local documentation: http://localhost:4000/sage/pages/objects#live_avatar

Screen Shot 2020-03-12 at 6 22 34 PM


app/assets/stylesheets/sage/core/_mixins.scss
```diff
@@ -72,3 +72,22 @@
 }
 
 
+/* ==================================================
+** Live Avatar Ring
+  Usage: `@include overflow-scroll();
```
teenwolfblitzer: You'll probably want to update this comment (or remove entirely, for now)

app/assets/stylesheets/sage/patterns/objects/_live_avatar.scss
```diff
@@ -0,0 +1,72 @@
...
+  &::before {
+    $sage-live-avatar-ring-offset: 0;
+
+    content: "";
+    position: absolute;
+    top: $sage-live-avatar-ring-offset;
+    right: $sage-live-avatar-ring-offset;
+    bottom: $sage-live-avatar-ring-offset;
+    left: $sage-live-avatar-ring-offset;
+    border-radius: 50%;
```
teenwolfblitzer: 📎 Can make use of `sage-border(radius-round)` here

app/assets/stylesheets/sage/patterns/objects/_live_avatar.scss
```diff
@@ -0,0 +1,72 @@
...
+@each $color in $sage-live-avatar-ring-colors {
+  .sage-live-avatar--ring-#{$color} {
+    @include sage-live-avatar-ring($color);
+  }
+}
+
+.sage-live-avatar__image {
+  max-width: 100%;
+  max-height: 100%;
+  border-radius: 50%;
```
teenwolfblitzer: `sage-border(radius-round)` here as well

app/views/sage/examples/objects/live_avatar/_preview.html.erb
```diff
@@ -0,0 +1,19 @@
+<!-- Showing no ring -->
+<div class="sage-live-avatar">
+  <img src="/assets/sage/avatar/jay.png" class="sage-live-avatar__image" alt="User's name"/>
```
teenwolfblitzer, line 3: Minor nitpick but the closing `/` for these is invalid HTML5 😈 :pitchfork:

app/assets/stylesheets/sage/patterns/objects/_live_avatar.scss
```diff
@@ -0,0 +1,72 @@
+/* ==================================================
+  ** live_avatar
+  type: object
+
+================================================== */
+
+$sage-live-avatar-size: 128px;
```
teenwolfblitzer: I think we'll want `8rem` for relative unit sizing here.

This also reminds me that I should create a px-to-rem mixin so we no longer need to calculate values outright.

app/assets/stylesheets/sage/patterns/objects/_live_avatar.scss
```diff
@@ -0,0 +1,72 @@
...
+$sage-live-avatar-ring-transition: $sage-btn-transition;
```
teenwolfblitzer: Phil hit upon a great idea of using the prefix `$-<module name>` for locally scoped variables. This should help 
 distinguish the global `$sage-` variables from local. For example, we'd have `$-live-avatar-ring-transition` here.

teenwolfblitzer (approved):
👍 This is looking 🔥, a few minor items to consider.

app/views/sage/examples/objects/live_avatar/_preview.html.erb
philschanely: 🤔 Invalid seems a bit harsh :pitchfork: haha. My understanding is that is merely not required but still a good practice. While XHTML is no longer enforced I appreciated the clarity of using either opening AND closing tags OR a distinct self-closing tag with the `/`. This makes it super-clear when you have a standalone tag rather than an accidental stray opening tag with no closer. Our tooling is smart enough today that that may not be necessary, but its a syntax I do prefer. https://dev.w3.org/html5/html-author/#self-closing-tag

If on the other hand you're suggesting that we follow a style that always uses "opening" tag syntax for void elements and no `/`, that's a whole different matter and I'll happily concede.

app/assets/stylesheets/sage/patterns/objects/_live_avatar.scss
philschanely: I roughed one in here. If you like it, @monkeypox8 it will henceforth be available to integrate in lieu of explicit `rem` settings when desired. Improves readability of `rem` settings IMO :) (since they're a little abstract, and using an integer value is easier to sync up with specs from mockups)

app/assets/stylesheets/sage/patterns/objects/_live_avatar.scss
philschanely: Thanks, @monkeypox8

app/assets/stylesheets/sage/patterns/objects/_live_avatar.scss
philschanely: Done! Thanks 👍

app/assets/stylesheets/sage/patterns/objects/_live_avatar.scss
philschanely: Done! Thanks 👍

PR #82, opened 2020-03-13
Form inputs

teenwolfblitzer:
Form inputs should be styled to match Figma spec.

- Default text inputs
- Specialized text inputs (email, password, ??)
- ~Textareas~ will apply separate element/treatment TBD

example:
input-fields

spec:
form inputs

## Related issues
- #92 stylelint update

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -141,6 +141,7 @@
     "selector-class-pattern": null,
     "scss/double-slash-comment-whitespace-inside": "always",
     "scss/dimension-no-non-numeric-values": true,
+    "scss/dollar-variable-pattern": "^[-]?[a-z]+([a-z0-9-]+[a-z0-9]+)?$",
     "scss/operator-no-unspaced": true,
     "scss/no-duplicate-mixins": true,
     "selector-no-qualifying-type": [true, { "ignore": ["attribute"] }],
--- a/app/assets/stylesheets/sage/patterns/elements/_form_input.scss
+++ b/app/assets/stylesheets/sage/patterns/elements/_form_input.scss
@@ -4,11 +4,6 @@
 
 ================================================== */
 
-@mixin floating-label {
-  transform: translateY(calc(-50% - (#{$sage-inputfield-font-size} - #{$sage-inputfield-padding-filled})));
-  opacity: 1;
-}
-
 .sage-input {
   position: relative;
   height: $sage-inputfield-height;
@@ -20,7 +15,7 @@
 
 .sage-input__label {
   position: absolute;
-  transform: translateY(calc(-1 * #{$sage-inputfield-font-size}));
+  transform: translateY(calc(#{$sage-inputfield-font-size} * -1));
   left: $sage-inputfield-padding;
   top: 50%;
   color: inherit;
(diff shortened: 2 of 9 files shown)


.stylelintrc.json
```diff
@@ -141,10 +141,11 @@
     "selector-class-pattern": null,
     "scss/double-slash-comment-whitespace-inside": "always",
     "scss/dimension-no-non-numeric-values": true,
+    "scss/dollar-variable-pattern": "^[-]?[a-z]+([a-z0-9-]+[a-z0-9]+)?$",
```
AndrwM, line 144: Very cool, I didn't realize you could check variable patterns against a regex.

app/assets/stylesheets/sage/patterns/elements/_form_input.scss
```diff
@@ -0,0 +1,99 @@
+/* ==================================================
+  ** _form_inputs.scss
+  type: element
+
+================================================== */
+
+@mixin floating-label {
+  transform: translateY(calc(-50% - (#{$sage-inputfield-font-size} - #{$sage-inputfield-padding-filled})));
+  opacity: 1;
+}
```
AndrwM: Should this be scoped so that it isn't accessible from outside this element/file?

app/assets/stylesheets/sage/patterns/elements/_form_input.scss
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  &:required:not(:placeholder-shown):not(:valid) {
+    border-color: $sage-inputfield-color-error;
+    box-shadow: $sage-inputfield-box-shadow-size $sage-inputfield-color-error;
+
+    ~ .sage-input__label {
+      color: $sage-inputfield-color-error;
+    }
+  }
```
AndrwM, line 103: It would be a good idea to explicitly state in the docs that the `<label>` _must_ come after the `<input>` for this component to operate as expected.

app/assets/stylesheets/sage/patterns/elements/_form_input.scss
```diff
@@ -0,0 +1,99 @@
...
+  &:active:not(:disabled) {
+    border-color: currentColor;
+    outline: none;
+
+    &::placeholder {
+      opacity: 0;
+    }
+  }
+
+  &:valid:not(:placeholder-shown) {
```
AndrwM, line 87: Is this `:not()` implementation IE11 compatible?

At first glance its a bit mooshy if this is supported: https://caniuse.com/#search=%3Anot

Just want to throw a flag to make sure we're considering a fallback if there are support issues.

AndrwM (approved):
The sliding label is hands down my favorite input UX, really excited about this.

Left a few notes.

app/assets/stylesheets/sage/patterns/elements/_form_input.scss
teenwolfblitzer: Great callout, I'll make that addition in the docs 👍

philschanely (approved):
Beautiful! 🏆

cmcfadzean (approved):
nice

app/assets/stylesheets/sage/patterns/elements/_form_input.scss
teenwolfblitzer: the `:not()` selector is IE11 compatible, but the `:placeholder-shown` is not. So the CSS validation colors don't work with IE until the form is submitted, which means we'll probably have to follow-up with classes once an actual form has been wired up. For now, I've separated out those declarations so that IE can at least display the focus and active states correctly.

PR #91, opened 2020-03-13
Add live profile card

### LINKED ISSUE #94: Add Live User Control

philschanely:
Resolves #94 

- [x] Add live user control object
- [x] Add live profile card
- [x] Adjust avatar for IE 11 
- [x] Adjust avatar to show Initials and colors
- [x] Implement Sage icons

#### Extra

- [x] Fix icon regression and semantics on quicklinks

### Testing

Open the Sage application. Navigate to Objects and then Live Proflie Card and check against design specs in [design mockup](https://www.figma.com/file/ikwcMcE6HEb653b0oVCgy6/%E2%9A%99%EF%B8%8FKajabi-Live-SAGE?node-id=15%3A142)


camsimony (approved):

philschanely:
Sorry this is such a large PR, y'all. Its partially due to having a number of subcomponents that benefited from being developed together and partially because I reworked the Quick Links component due to an icon and semantics issue.

app/assets/stylesheets/sage/core/_mixins.scss
```diff
@@ -95,6 +95,46 @@
...
+  arguments: $icon | null|[string]
+================================================== */
+
+@mixin icon-base($icon: null) {
+  display: inline-block;
+  text-transform: none;
+  font: normal normal normal 1rem/1 "Sage";
+  speak: none;
+  /* Better Font Rendering */
+  -webkit-font-smoothing: antialiased;
```
teenwolfblitzer: I think we're safe to remove this now, since we're already enabling it on the `body`.

app/assets/stylesheets/sage/core/_variables.scss
```diff
@@ -73,6 +73,8 @@ $sage-btn-focus-outline-color: currentColor !default;
 
 // Primary button
 $sage-btn-primary-bg-color: sage-color(primary) !default;
+$sage-btn-primary-bg-active-color: darken($sage-btn-primary-bg-color, 11%);
```
teenwolfblitzer: Should probably add the `!default` flag here (and the following line as well)

app/assets/stylesheets/sage/patterns/elements/_button.scss
```diff
@@ -39,11 +39,13 @@
   }
 }
 
-
+// TODO: Re-evaluate whether these need to be "on" these elemets by default.
```
teenwolfblitzer: :neckbeard:  `elements`

app/assets/stylesheets/sage/patterns/objects/_live_avatar.scss
```diff
@@ -36,34 +34,74 @@ $-live-avatar-ring-colors: (
...
 }
 
+// Documentation-specific styles
+.sage-live-avatar-wrapper {
+  display: flex;
+  margin-bottom: sage-spacing();
+}
+
 .sage-live-avatar--ring {
   @include sage-live-avatar-ring();
```
teenwolfblitzer: The naming here feels like it could cause confusion between `.sage-live-avatar--ring` and `sage-live-avatar-ring`. I'm wondering if we should start to standardize mixin naming with `sage-mixin-`, or just `mixin-` to make it evident. Can go in a follow-up once we decide on direction.

app/assets/stylesheets/sage/core/_mixins.scss
```diff
@@ -105,12 +145,15 @@
 
 @mixin sage-live-avatar-ring($color: primary) {
```
teenwolfblitzer: See my note above on mixin naming conventions.

app/assets/stylesheets/sage/patterns/objects/_live_avatar.scss
```diff
@@ -36,34 +34,74 @@ $-live-avatar-ring-colors: (
...
+  }
+
+  // Default initials
+  &::after {
+    content: attr(data-initials);
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    position: absolute;
+    z-index: 1;
```
teenwolfblitzer: Can we make use of the `sage-z-index` token here?

app/assets/stylesheets/sage/patterns/objects/_live_option_menu.scss
```diff
@@ -0,0 +1,127 @@
+/* ==================================================
+  ** live_option_menu
+  type: object
+
+
+================================================== */
+
+$-live-option-menu-button-height: 40px;
```
teenwolfblitzer: Should these values be using the `rem()` function?

app/assets/stylesheets/sage/core/_mixins.scss
```diff
@@ -95,6 +95,46 @@
...
 
+/* ==================================================
+** Absolute Positioning shorthand
+  Usage: `@include position-full(4px);`
+
+  Sets an element's positioning offsets all to the value provided.
+  arguments: $offset | [length]
+================================================== */
+
+@mixin position-full($offset: 0) {
```
teenwolfblitzer: 😎

app/assets/stylesheets/sage/patterns/objects/_live_avatar.scss
```diff
@@ -36,34 +34,74 @@ $-live-avatar-ring-colors: (
...
+
+    &.sage-live-avatar--ring {
+      @include sage-live-avatar-ring($color);
+    }
   }
 }
 
 .sage-live-avatar__image {
   position: relative;
+  z-index: 2;
```
teenwolfblitzer: Can we make use of the `sage-z-index` token here?

app/assets/stylesheets/sage/patterns/objects/_live_option_menu.scss
```diff
@@ -0,0 +1,127 @@
...
+    $-live-option-menu-focus-offset-inline: 50%;
+    content: "";
+    display: block;
+    position: absolute;
+    right: $-live-option-menu-focus-offset-inline;
+    bottom: 0;
+    left: $-live-option-menu-focus-offset-inline;
+    border-bottom: 2px solid transparent;
+    transition: 0.1s ease-in-out;
+    transition-property: border-bottom-color, left, right;
```
teenwolfblitzer: 🔥 🥇I like this! Feels a lot cleaner than the way I've been writing transitions as single-line, comma-separated per property

app/assets/stylesheets/sage/patterns/objects/_live_profile_card.scss
```diff
@@ -0,0 +1,58 @@
+/* ==================================================
+  ** live_profile_card
+  type: object
+
+
+================================================== */
+
+$-live-profile-card-avatar-size: 38px;
```
teenwolfblitzer: Should this value be using the `rem()` function?

app/assets/stylesheets/sage/patterns/objects/_live_profile_card.scss
```diff
@@ -0,0 +1,58 @@
...
+  margin-bottom: sage-spacing(2xs);
+}
+
+.sage-live-profile-card__controls {
+  display: flex;
+  margin-left: auto;
+}
+
+.sage-live-profile-card__info {
+  margin-right: 16px;
```
teenwolfblitzer: Should be able to use `sage-spacing` here

app/assets/stylesheets/sage/patterns/objects/_live_user_control.scss
```diff
@@ -0,0 +1,69 @@
+/* ==================================================
+  ** live_user_control
+  type: object
+
+
+================================================== */
+
+$-live-user-control-size: 36px;
```
teenwolfblitzer: Should this value (and the following) be using the `rem()` function?

app/assets/stylesheets/sage/tokens/_color_palette.scss
```diff
@@ -60,4 +60,11 @@ $sage-colors: (
     400: #304050,
     500: #263240,
   ),
+  white: (
+    100: #fff,
+    200: #fff,
```
teenwolfblitzer: Curious about this one… are the additional white values for potential shades?

app/assets/stylesheets/sage/tokens/_spacing.scss
```diff
@@ -4,6 +4,7 @@
 }
 
 $sage-spacings: (
+  3xs: 0.125rem, // 2px
```
teenwolfblitzer: 🤔 Makes me wonder if we should adjust our scale by one or two levels, making `8px` or `16px` the base instead.

app/views/sage/pages/color.html.erb
```diff
@@ -6,11 +6,8 @@
 <% end %>
 <%= content_for :quick_links do %>
   <div class="quick-links">
-    <div class="quick-links__header">
-      <i class="quick-links__icon sage-icon sage-icon-file-content" aria-hidden="true"></i>
-      <h5 class="quick-links__title">Contents</h5>
-    </div>
-    <ul class="quick-links__body">
+    <h2 class="quick-links__title t-sage-heading-6">Contents</h2>
```
teenwolfblitzer, line 9: 💥

app/assets/stylesheets/sage/patterns/objects/_live_user_control.scss
```diff
@@ -0,0 +1,69 @@
...
+$-live-user-control-icon-size: 24px;
+$-live-user-control-focus-offset: 0;
+$-live-user-control-focus-width: $sage-btn-focus-outline-width * 1px;
+$-live-user-control-color: $sage-btn-primary-bg-color;
+$-live-user-control-color-hover: $sage-btn-primary-bg-hover-color;
+$-live-user-control-inactive-color: sage-color(grey, 400);
+$-live-user-control-inactive-color-hover: sage-color(grey, 500);
+
+.sage-live-user-control {
```
teenwolfblitzer: I'm seeing the default button background showing up in Firefox, Safari, and Edge. Pretty sure it's a result of this class needing `background: transparent` since it's not using `button-style-reset()`:

Safari
safari-button-bkgd

I'm also running into some odd behavior with the option menu button not appearing at all in Edge 44/EdgeHTML 18:
edge-button-hover

app/assets/stylesheets/sage/patterns/objects/_live_option_menu.scss
```diff
@@ -0,0 +1,127 @@
...
+
+    @include position-full(0);
+  }
+
+  // Specific to preview
+  .sage-docs .example__preview > & {
+    display: flex;
+    align-items: center;
+    justify-content: center;
+    width: 48px;
```
teenwolfblitzer: Here as well: should these be using the `rem()` function?

teenwolfblitzer (commented):
Looking good, couple of questions around the profile buttons and some minor suggestions to consider. Nice work!

app/assets/stylesheets/sage/patterns/objects/_live_option_menu.scss
philschanely: Got it.

app/views/sage/pages/color.html.erb
philschanely: I find my self on the fence a lot with this too. Definitely not arguing with the modularization of the type specs but it does seem a bit inefficient thinking about having to put this on every implementation of the component. It makes most sense to me for those that may change specs frequently or depending on implementation context, but I'm finding that for more "settled" components that extending the `%` form of the spec accomplishes similar goals of isolating such type settings. Thoughts @monkeypox8 ? You can see this in the other PR I have up: https://github.com/Kajabi/sage/pull/115

app/assets/stylesheets/sage/tokens/_spacing.scss
philschanely: Or perhaps choose that using it for a 2px spacer is perhaps pushing it too far? ;) I'll undo this for now. Actually didn't even use it after all haha

app/assets/stylesheets/sage/tokens/_color_palette.scss
philschanely: Yeah seems its most convenient to use white with the same sage-color token function, and rather than add a condition to the function to just return one value when white is requested, I opted to just add a stacked setting here. Thoughts @monkeypox8 ?

app/assets/stylesheets/sage/core/_mixins.scss
philschanely: Thanks!

app/assets/stylesheets/sage/patterns/objects/_live_avatar.scss
philschanely: Ah! Of course!

app/assets/stylesheets/sage/patterns/objects/_live_option_menu.scss
philschanely: Sure! In fact I'm going to suggest that with `strip-unit` in place in the `rem` function we should keep the units in the value we pass it in order to make the intention cleaner. Also if we ever back out of such rampant use of `rem` we have the desired units there for easier replacing.

app/assets/stylesheets/sage/patterns/objects/_live_avatar.scss
philschanely: yes!

app/assets/stylesheets/sage/patterns/objects/_live_option_menu.scss
philschanely: Thanks! I forgot to use the `$sage-transition` variable though... will implement.

app/assets/stylesheets/sage/patterns/elements/_button.scss
philschanely: 😊

app/assets/stylesheets/sage/patterns/objects/_live_avatar.scss
philschanely: I see your point. I updated to `sage-feature-live-avatar-ring` for now; if another convention is preferred we can switch later.

app/assets/stylesheets/sage/patterns/objects/_live_user_control.scss
philschanely: Thanks, @monkeypox8 . I fixed the button background issue. On the menu bug on one hand only the first menu is set up (lots of extra markup otherwise) but that first one should work at least. On the other hand, the code here is only meant for demonstration purpose, as the actual component will be built in React in Products. So not sure how big of an issue to make of it. Could we perhaps pair on this to look together?

teenwolfblitzer: Per our call, turns out that the Edge issue was caused by grouping `:focus-within` with the `:hover` pseudo-selector. Separating the two should amend this, as Edge and IE11 will ignore the entire ruleset when it encounters `:focus-within`.

teenwolfblitzer (commented):
One more item to check out: in `_icons.scss`, the new asset pipeline has prevented the icon font from loading, so the `url()` references in line 16 should be changed to `asset-url()`.

teenwolfblitzer (approved):
👍 🔥 Looks good to me, nice work!

PR #90, opened 2020-03-13
Move JS source files

cmcfadzean:
JS is currently located in `test/dummy/app/javascripts/` and should be moved out to the overall `app/assets` directory


teenwolfblitzer:
Oh interesting, so we don't need to move `application.js` and `pages.js` along with `docs.js`? I was under the impression I'd need to change all the references for future JS modules.

teenwolfblitzer (commented):
👍 Looks good to me!

PR #103, opened 2020-03-16
Add live active mic icon

### LINKED ISSUE #102: Active Microphone Icon
Add an animatable active microphone icon.

Design spec to come.

philschanely:
Resolves #102 

NOTE: This implementation shows the functionality on "click" but the actual implementation int he Live app will work together with live audio data. 

- [x] Mock up animating microphone icon
- [x] Sync with design


pixelflips (approved):
👍 Nice work! It looks to be working as expected.

app/assets/stylesheets/sage/patterns/objects/_live_active_mic.scss
```diff
@@ -0,0 +1,55 @@
+/* ==================================================
+  ** live_active_mic
+  type: object
+
+
+================================================== */
+
+.sage-live-active-mic {
+  --color-volume: #{sage-color(sage)};
```
teenwolfblitzer: Should there be a fallback value for the custom props with IE11? The image does display as a static icon just fine (no animation) in IE11, so it's probably reasonable to treat the animation as an "enhancement".

teenwolfblitzer (approved):
👍 😎 Looks great! Left one question about IE11 support that may be optional.

app/assets/stylesheets/sage/patterns/objects/_live_active_mic.scss
philschanely: As "Sexy" as this approach is perhaps I should just switch to using direct properties. 🤔

AndrwM (approved):
This is nice! +1

philschanely:
Y'all, I decided to back out from a good bit of the custom prop stuff, leaving only the volume value. And that itself is simpler now as well to make for animating the value much simpler in an IE-11 compatible environment. Docs are updated too. Lets land this!

PR #56, opened 2020-03-17
Typography style refactor

teenwolfblitzer:
## Description
Headlines and overall typography are currently styled by their element names (`h1, h2, h3`, etc). By using classes for styling these elements, we'll be better able to maintain semantic HTML separate from styling, and prevent overrides of unintended margins, padding, etc.

- [x] Restructure heading styles to use classes
- [x] Create `letter-spacing` token
- [x] Globalize base type styles for theming
- [x] Update styles throughout documentation to use new classnames
- [x] Modify documentation in Sage around type and its use
- [x] Add documentation in Guidelines regarding adding type specs

### Steps for testing
No visual changes should be introduced with this refactor.
1. Start Sage locally
2. Navigate to the [Typography](http://localhost:4000/sage/pages/typography) overview page
3. Compare styles with Figma to verify spec remains the same

[Documentation in Guidelines](https://github.com/Kajabi/sage/wiki/Understanding-and-Maintaining-Type-Specs)


app/assets/stylesheets/sage/core/_functions.scss
```diff
@@ -14,3 +14,22 @@
 @function rem($value) {
   @return $value / 16 * 1rem;
 }
+
+
+/* ==================================================
+  ** Strip units from value
+  Usage: `rem(12)`
+
+  Converts an integer to a rem value based on 16 as the base size.
```
philschanely, line 23: Does this documentation need to be synced with the actual function below?

app/assets/stylesheets/sage/core/_typography.scss
```diff
@@ -3,88 +3,103 @@
 
 ================================================== */
 
+// ==================================================
```
philschanely, line 6: Consider `t-body-block-spacing` to clarify this mixin's use further.

app/views/sage/pages/index.html.erb
```diff
@@ -1,10 +1,10 @@
 <%= content_for :heading do %>
-  <h1>Welcome to Sage</h1>
+  <h1 class="t-heading-1">Welcome to Sage</h1>
   <p>The Sage Design System (SDS) is our single source of truth, providing everything you need to build great products for our customers. It is the culmination of designers and developers working together to give teams the ability to ship high-quality products faster.</p>
 <% end %>
 <div class="sage-panel">
   <%= image_tag("sage/sage_illustration.png", class: "sage-panel__img")%>
-  <h3 id="how_works">How it works</h3>
+  <h3 class="t-heading-3" id="how_works">How it works</h3>
```
philschanely: Is this an example of a place we should switch the tag to `h2` but use the `heading-3` class for styling?

app/assets/stylesheets/sage/core/_typography.scss
```diff
@@ -3,88 +3,103 @@
...
-  margin-bottom: sage-spacing();
   text-rendering: optimizeLegibility;
   -webkit-font-smoothing: antialiased;
   -moz-osx-font-smoothing: grayscale;
   text-size-adjust: 100%;
-  color: sage-color(charcoal, 400);
-  font-family: "Inter", $sage-font-stack;
-  font-weight: sage-font-weight();
-  font-size: sage-font-size();
+  color: $sage-body-font-color;
```
philschanely, line 144: Cool, so if I follow we're basically putting on the `body text` style here as the default everywhere, so `p`, `li`, etc can just inherit this. Nice solution for avoiding `t-body` everywhere. Should we perhaps have a mixing that reverses this we can use for times when we want such a simple element to us other styles?

philschanely (approved):
Small suggestions.

app/assets/stylesheets/sage/core/_functions.scss
teenwolfblitzer: Oh whoops! Great catch.

teenwolfblitzer:
May be tabling this one temporarily for a more thorough rework.

app/assets/stylesheets/sage/core/_typography.scss
```diff
@@ -3,100 +3,215 @@
...
+  "body-xsmall-bold": map-merge($t-body-xsmall-spec, (
+    weight: sage-font-weight(bold)
+  )),
+
+  // Additional custom specs
+  // TBD
+);
+
+// Automate setting up base specs to extend in context.
+@each $spec-name, $spec in $type-specs {
```
teenwolfblitzer, line 108: 👏

app/assets/stylesheets/sage/core/_typography.scss
```diff
@@ -3,100 +3,215 @@
...
-    text-decoration: none;
+    &:hover,
+    &:focus {
+      color: inherit;
+      text-decoration: none;
+    }
   }
 }
 
+// TODO: Are these used?
```
teenwolfblitzer: I don't think we have a style for `hr` that I've seen yet, so it can be removed.

app/assets/stylesheets/sage/core/_typography.scss
```diff
@@ -3,100 +3,215 @@
...
+    color: sage-color(primary);
 
-  &:hover,
-  &:focus {
-    color: inherit;
-    text-decoration: none;
+    &:hover,
+    &:focus {
+      color: inherit;
+      text-decoration: none;
```
teenwolfblitzer: Something to note, the sidenav now has underlined links on hover because of the added specificity here. I wouldn't go so far as to say it's a regression, since we don't have a defined style for the sidenav links, let alone text links and hover/focus states. Needs a follow-up once we sync with design on what that both of those should look like.

teenwolfblitzer (commented):
👍 Since I originated this PR, I'm not able to approve, but it's really looking great; love how efficient you've made it. Had one non-blocking comment to investigate regarding the sidenav and text links in general.

app/assets/stylesheets/sage/core/_typography.scss
philschanely: 🙇

app/assets/stylesheets/sage/core/_typography.scss
philschanely: Ooo good catch. I'll look into it.

philschanely: Handled.

app/assets/stylesheets/sage/core/_typography.scss
philschanely: I moved and clarified the comment. I see some `hr` used but feel future work should consider consolidating the `label` styles.

AndrwM (approved):
This is really clever. I like how you executed the `.sage-type` when naked type classes exist. +100

philschanely:
Filed #113 for integrating the other aspects of the existing repo with these type changes and landing this one. Thanks, all!

PR #83, opened 2020-03-17
Tooltips

pixelflips:
Create and style tooltip as defined in Figma
tooltip


pixelflips:
@cmcfadzean - I still have some styling and clean up to take care of, but could you please have a look and double-check the js structure we discussed.

config/initializers/assets.rb
```diff
@@ -8,4 +8,5 @@
 
 # Precompile additional assets.
 # application.js, application.css, and all non-JS/CSS in app/assets folder are already added.
-Rails.application.config.assets.precompile += %w( docs.js )
\ No newline at end of file
+Rails.application.config.assets.precompile += %w( docs.js )
```
cmcfadzean: I believe that this should be sage_docs.js

app/views/sage/examples/elements/tooltip/_props.html.erb
```diff
@@ -0,0 +1,6 @@
+<tr>
```
cmcfadzean, line 1: Might be good to fill in some of this info that we expect to have such as position and theme

app/helpers/sage/elements_helper.rb
```diff
@@ -13,6 +13,19 @@ module ElementsHelper
     def sage_elements
       [
         # Sage Generated Elements
+        {
+          title: "tooltip",
+          description: "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.", 
+          scss_design:  "todo",
```
cmcfadzean: done?

app/helpers/sage/elements_helper.rb
```diff
@@ -13,6 +13,19 @@ module ElementsHelper
...
+          scss_dev:     "todo",
```
cmcfadzean: doing?

app/assets/stylesheets/sage/patterns/elements/_tooltip.scss
```diff
@@ -0,0 +1,81 @@
+/* ==================================================
+  ** tooltip
+  type: element
+
+
+
+================================================== */
+
+$tooltip-class: "sage-tooltip" !default;
```
cmcfadzean: I think now we are putting these variables up in the global file.

cmcfadzean (commented):
General feedback on process and structure. Looking forward to the other UXD's to go deep on the code. Looks great tho!

config/initializers/assets.rb
pixelflips: thanks!

app/assets/stylesheets/sage/patterns/elements/_tooltip.scss
pixelflips: 👍

app/helpers/sage/elements_helper.rb
pixelflips: 👍

app/helpers/sage/elements_helper.rb
pixelflips: updated to "doing". will adjust after review.

app/views/sage/examples/elements/tooltip/_props.html.erb
pixelflips: 👍

philschanely (approved):
Nice work here!

cmcfadzean (approved):
Nice

app/assets/javascripts/tooltip.js
```diff
@@ -0,0 +1,66 @@
+Tooltip = function(options) {
```
AndrwM, line 1: It might be best to avoid future collisions with other js to next this under a global variable, possibly – `window.sage.tooltp`

app/assets/javascripts/tooltip.js
```diff
@@ -0,0 +1,66 @@
...
+      positionTooltip(e.target, tooltip, pos);
+    });
+  
+    document.body.addEventListener("mouseout", function(e) {
+      if (e.target.hasAttribute("data-tooltip")) {
+        setTimeout(function() {
+          document.body.removeChild(document.querySelector(".sage-tooltip"));
+        }, 0);
+      }
+    });
```
AndrwM, line 28: This will run whenever any `mouseout` event happens.
It would be best to bind these listeners directly to elements that have `data-tooltip`.

app/assets/javascripts/tooltip.js
```diff
@@ -0,0 +1,66 @@
+Tooltip = function(options) {
+    var theme =  dist = options.distance || 10;
```
AndrwM, line 2: I'm confused why both `dist` and `theme` are being assigned to the distance. Could you explain?

app/assets/javascripts/tooltip.js
```diff
@@ -0,0 +1,66 @@
...
+      var tooltip = document.createElement("div");
+      var pos = e.target.getAttribute("data-position") || "top";
+      tooltip.className = "sage-tooltip";
+      tooltip.innerHTML = e.target.getAttribute("data-tooltip");
+      tooltip.theme = e.target.getAttribute("data-tooltip-theme");
+      tooltip.position = e.target.getAttribute("data-position");
+
+      if (tooltip.theme) {
+        tooltip.classList.add("sage-tooltip--light");
+      }
```
AndrwM, line 16: It'd be best to be able to name the theme in the styles and then pass that name through via the `data-tooltip-theme`  attribute.

So that we can define:
```
.sage-tooltip--NAME {}
```
that is then referred to in the html:
```
<div data-tooltip-theme="NAME" />
```

app/assets/javascripts/tooltip.js
```diff
@@ -0,0 +1,66 @@
...
+    document.body.addEventListener("mouseout", function(e) {
+      if (e.target.hasAttribute("data-tooltip")) {
+        setTimeout(function() {
+          document.body.removeChild(document.querySelector(".sage-tooltip"));
+        }, 0);
+      }
+    });
+  
+    function positionTooltip(parent, tooltip, position) {
+      var parentCoords = parent.getBoundingClientRect(), left, top;
```
AndrwM, line 31: This is cool, I didn't realize you could define `var`s like this.

AndrwM (commented):
This is great, I left a few notes

AndrwM:
Awh, looks like I was late to the party on the review :\

pixelflips:
> Awh, looks like I was late to the party on the review :\

Not a problem at all. Those are very helpful and I can create a new PR for some JS cleanup. Thanks for taking the time!

PR #96, opened 2020-03-19
Text areas

teenwolfblitzer:
Text area styling should match Figma spec:

textarea

`:focus` and validation states should match input areas, where applicable.


app/assets/stylesheets/sage/core/_variables.scss
```diff
@@ -86,10 +86,11 @@ $sage-inputfield-font-size: $sage-body-font-size !default;
 $sage-inputfield-letter-spacing: $sage-body-letter-spacing !default;
 
 // Floating label
-$sage-inputfield-label-font-size: 0.6875rem !default;
+$sage-inputfield-label-font-size: 0.6875rem !default; // 11px
```
philschanely: Could we start working in the `rem()` function to avoid needing the comment at the end showing the original units?

app/assets/stylesheets/sage/core/_variables.scss
```diff
@@ -101,17 +102,59 @@ $sage-inputfield-padding-offset: $sage-inputfield-padding - $sage-inputfield-bor
...
 $sage-inputfield-color-error: sage-color(red) !default;
 $sage-inputfield-bg-disabled: sage-color(grey, 100) !default;
 
 
+// ==================================================
+// TEXTAREAS
+// ==================================================
+
+$sage-textarea-font-size: $sage-inputfield-font-size !default;
+$sage-textarea-line-height: calc(30 / 16) !default;
```
philschanely: Can we use one of our line-height tokens here?

app/assets/stylesheets/sage/core/_variables.scss
```diff
@@ -101,17 +102,59 @@ $sage-inputfield-padding-offset: $sage-inputfield-padding - $sage-inputfield-bor
...
+$sage-textarea-label-line-height: 1 !default;
+
+// Border
+$sage-textarea-border-color: $sage-inputfield-border-color !default;
+$sage-textarea-border-radius: $sage-inputfield-border-radius !default;
+$sage-textarea-border-width: $sage-inputfield-border-width !default;
+$sage-textarea-box-shadow-size: $sage-inputfield-box-shadow-size !default;
+
+// Padding
+$sage-textarea-label-padding-top: 0.75rem !default; // 12px
```
philschanely: More places for `rem()`...

app/assets/stylesheets/sage/patterns/elements/_form_textarea.scss
```diff
@@ -0,0 +1,106 @@
...
+  &:not(:last-child) {
+    margin-bottom: $sage-textarea-spacing;
+  }
+}
+
+.sage-textarea__label {
+  position: absolute;
+  left: $sage-textarea-padding;
+  top: $sage-textarea-border-width;
+  width: $sage-textarea-label-width;
```
philschanely: Do you see this changing from 100%? I wonder if we could use `right: $sage-textarea-padding` instead?

app/views/sage/examples/elements/form_textarea/_preview.html.erb
```diff
@@ -0,0 +1,16 @@
+<p>Default</p>
+<div class="sage-textarea">
```
philschanely, line 2: a11y question: Does the `for` pairing mean that the field's label gets read regardless of its proximity? I would have thought that it would be best for the label to precede the field. 🤔

philschanely (commented):
A few thoughts before 👍

app/assets/stylesheets/sage/core/_variables.scss
teenwolfblitzer: 👍 That's a great point. I'll also open an issue to sweep through and change existing values in other files as well.

app/views/sage/examples/elements/form_textarea/_preview.html.erb
teenwolfblitzer: Yes, it gets read regardless. I do think we'll have to do additional reader testing, but working through Voiceover in Safari appeared to be flowing correctly, though with an additional focus state once the label was read. The floating label pattern would require JS to function correctly in the opposite direction (label preceding input), but an alternative may be to wrap the entire field in a label.

app/assets/stylesheets/sage/patterns/elements/_form_textarea.scss
teenwolfblitzer: Hm, I was planning to leave the door open for varying textarea widths, since I see this potentially being used in a dialog or split-view form. I'll check with design and create a follow-up patch if needed.

app/assets/stylesheets/sage/core/_variables.scss
teenwolfblitzer: 🙌 Good call; I see one other place where this line-height may be used, so I've added a new token.

philschanely (approved):

PR #110, opened 2020-03-21
Realign switch styles

teenwolfblitzer:
Focus and hover states for the switch element vary slightly from the Figma spec as seen in the attached screen.
- Focus state outline color is sage/grey instead of sage/primary
- Hover state increases intensity of the drop shadow on the toggle control and switch body
- Add error state styling

Expected:
switch

Actual: (focus)
switch-focus-current

### DIFF
--- a/app/views/sage/examples/elements/switch/_preview.html.erb
+++ b/app/views/sage/examples/elements/switch/_preview.html.erb
@@ -3,18 +3,18 @@
     <!-- Default, checkbox -->
     <div class="sage-switch">
       <input type="checkbox" class="sage-switch__input" id="sage-switch-1" name="sage-switch-1" value="switch-1-value">
-      <label for="sage-switch-1">Switch (checkbox)</label>
+      <label class="sage-switch__label" for="sage-switch-1">Switch (checkbox)</label>
     </div>
   </div>
   <div class="sage-col--md-6">
     <!-- Default, radio -->
     <div class="sage-switch">
       <input type="radio" class="sage-switch__input" id="sage-switch-2a" name="sage-switch-2" value="switch-2a-value" checked>
-      <label for="sage-switch-2a">Switch (radio)</label>
+      <label class="sage-switch__label" for="sage-switch-2a">Switch (radio)</label>
     </div>
     <div class="sage-switch">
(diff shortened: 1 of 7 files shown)


cmcfadzean (approved):
Like the class adds

app/assets/stylesheets/sage/patterns/elements/_switch.scss
```diff
@@ -7,101 +7,111 @@
...
+  vertical-align: top;
+  appearance: none;
+  color: $sage-switch-color-default;
+  background: $sage-switch-color-default;
+  border-radius: $sage-switch-border-radius;
+  outline: none;
+  transition: background 0.3s ease-out;
+  cursor: pointer;
+
+  + label {
```
AndrwM: Marking this to ensure we're communicating in the docs that `<label>`s must come after `<input.sage-switch__input>` in the HTML.

philschanely: Could we implement a BEM setup on this as well?

AndrwM (approved):
Nice! +1

app/views/sage/examples/elements/switch/_preview.html.erb
```diff
@@ -1,15 +1,20 @@
 <!-- Default -->
 <div class="sage-switch">
-  <input id="s1" type="checkbox">
-  <label for="s1">Switch</label>
+  <input type="checkbox" class="sage-switch__input" id="sage-switch-1" name="sage-switch-1">
+  <label for="sage-switch-1">Switch</label>
```
philschanely: See BEM request above.

philschanely (commented):
Just requesting BEM be used on label. Great otherwise!

pixelflips (approved):
👍 Nice work. Just a thought and by no means a blocker at all, but I noticed in the Figma file the label is also red for the error switch. I'm not even sure it even applies since switches may not have a "label" in many cases.

teenwolfblitzer:
> 👍 Nice work. Just a thought and by no means a blocker at all, but I noticed in the Figma file the label is also red for the error switch. I'm not even sure it even applies since switches may not have a "label" in many cases.

Great callout, I've added the ability for the label to change to red as well. That being said, we'll need to do another pass for more useful form validation on the client side, since unchecked and `required` inputs here will default to appearing invalid on page load.

PR #114, opened 2020-03-24
adding gems

cmcfadzean:
Moves Gems so they will be dependencies of the gem. 
No longer requiring we add them to the parent app.

### DIFF
--- a/app/views/sage/application/_assistant.html.erb
+++ b/app/views/sage/application/_assistant.html.erb
@@ -5,10 +5,6 @@
   </button>
   <%= image_tag("sage/sage.svg", class: "sage-assistant__branding")%>
   <div class="sage-assistant__body">
-    <div class="sage-container--sm">
-      <label for="sage-assistant__search" class="visually-hidden">Search</label>
-      <input type="text" class="sage-assistant__search" id="sage-assistant__search" placeholder="Search&hellip;">
-    </div>
   </div>
   <div class="sage-assistant__actions"></div>
 </div>
--- a/Gemfile
+++ b/Gemfile
@@ -12,8 +12,3 @@ gemspec
 
 # To use a debugger
 # gem 'byebug', group: [:development, :test]
-
-gem 'sqlite3', '~> 1.3.0'
-gem 'sass-rails'
-gem 'autoprefixer-rails'
-gem "inline_svg"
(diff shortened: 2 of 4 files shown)


app/views/sage/application/_assistant.html.erb
```diff
@@ -5,10 +5,6 @@
   </button>
   <%= image_tag("sage/sage.svg", class: "sage-assistant__branding")%>
   <div class="sage-assistant__body">
-    <div class="sage-container--sm">
-      <label for="sage-assistant__search" class="visually-hidden">Search</label>
-      <input type="text" class="sage-assistant__search" id="sage-assistant__search" placeholder="Search&hellip;">
-    </div>
```
AndrwM, line 11: Why was this snippet removed?

AndrwM (approved):
Great! +1

app/views/sage/application/_assistant.html.erb
```diff
@@ -5,10 +5,6 @@
   </button>
   <%= image_tag("sage/sage.svg", class: "sage-assistant__branding")%>
   <div class="sage-assistant__body">
-    <div class="sage-container--sm">
-      <label for="sage-assistant__search" class="visually-hidden">Search</label>
```
pixelflips, line 9: Search was removed?

pixelflips (approved):
👍 Nice work!

philschanely (approved):

app/views/sage/application/_assistant.html.erb
cmcfadzean: The search was never hooked up its more of an in-app pattern so if we want to add search in later on we can do that as a full ticket but does not need to be cluttering the UI if it does not work currently.

app/views/sage/application/_assistant.html.erb
cmcfadzean: The search was never hooked up its more of an in-app pattern so if we want to add search in later on we can do that as a full ticket but does not need to be cluttering the UI if it does not work currently.

teenwolfblitzer (approved):
👍 :shipit: LGTM! Since Search was removed from the assistant, we should pull it out of the assistant example as well in a follow-up.

philschanely:
> 👍 :shipit: LGTM! Since Search was removed from the assistant, we should pull it out of the assistant example as well in a follow-up.

If its not in this app could it still perhaps be a feature in other contexts?

cmcfadzean:
Just a note on the search in the assistant. When we go to use this in the app there will be the search field but in the documentation its not used so it should not be in the search bar but should stay in the documentation of how the assistant should be used. Hope that makes sense.

PR #117, opened 2020-03-24
Jcm assets

cmcfadzean:
Precompile to inside gem so does not need to be called out in other parent apps. 
Follows better naming convention for files.


config/initializers/assets.rb
```diff
@@ -8,5 +8,10 @@
 
 # Precompile additional assets.
 # application.js, application.css, and all non-JS/CSS in app/assets folder are already added.
-Rails.application.config.assets.precompile += %w( sage_docs.js )
-Rails.application.config.assets.precompile += %w( sage.js )
+Rails.application.config.assets.precompile << /\.(?:svg|png|jpg|eot|woff|ttf)$/
```
teenwolfblitzer, line 11: Can we also add: `woff2`, `gif`, `mp4`, `webm` and `webp` to this list? I don't know if/when we'll ever use those last three but it'd cover our bases.

teenwolfblitzer (approved):
👍 Nice work! Will defer to you on the ruby updates, but compile is working as expected and the naming structure feels cleaner to me. One addition to consider on the precompiled assets for different extensions.

config/initializers/assets.rb
cmcfadzean: I think we can consider with a follow up.

PR #120, opened 2020-03-26
Fixes for prep to version bump

philschanely:
- [x] Fixing a few discrepancies between recent merges and the core for type styles.
- [x] Add white color to palette.
- [x] Adjusts JS away from ES6 syntax.


cmcfadzean:
Hold off on the version bump I need to fix an issue with assets

pixelflips (approved):
👍 +1 LGTM - thanks for the JS adjustments for tooltips!

PR #121, opened 2020-03-26
Viewport metatag missing for mobile devices

teenwolfblitzer:
Sage app HTML is missing the `<meta name="viewport" content="width=device-width, initial-scale=1">` meta tag. Because of this, mobile devices are rendering the Sage documentation page according to their screen resolution rather than by their device width.

This is reproducible on actual devices, and within the Xcode iOS simulator, but not by using Safari's "Responsive Design Mode".

- This PR also fixes an issue introduced from a missing variable in the textarea PR (#96 )

mobile-view


pixelflips (approved):
👍 +1 Looks good!

philschanely (approved):

PR #123, opened 2020-03-27
[mobile] Form field drop shadows

teenwolfblitzer:
## Description
Form inputs and text areas are displaying browser default drop shadows in iOS.

### Screenshots
|  Form inputs   |  Textareas  |
|--------|--------|
|before: ios-form-inputs|ios-form-textarea|
after: input-after|after: textarea-after|

## Related issues
n/a


cmcfadzean (approved):
I like reviewing this size PR lol

PR #126, opened 2020-03-27
[mobile] Switch element outline

teenwolfblitzer:
## Description
Switch element displays border around control in iOS:

### Screenshots
|  Before   |  After  |
|--------|--------|
|ios-switch|switch-after|


cmcfadzean (approved):
YES!

PR #101, opened 2020-03-27
Labels

pixelflips:
Match design and functionality for label/badge elements.

Spec:
label

### DIFF
--- a/app/assets/stylesheets/sage/system/patterns/elements/_label.scss
+++ b/app/assets/stylesheets/sage/system/patterns/elements/_label.scss
@@ -5,82 +5,58 @@
 
 ================================================== */
 
-$sage-label-padding: 0 sage-spacing(2xs);
-$sage-label-font-size: 75%;
-$sage-label-border-radius: sage-border(radius);
-
-$sage-label-draft-color: sage-color(charcoal, 200);
-$sage-label-draft-background: sage-color(grey, 200);
-$sage-label-draft-border-color: sage-color(grey, 200);
-
-$sage-label-success-color: sage-color(sage, 400);
-$sage-label-success-background: sage-color(sage, 100);
-$sage-label-success-border-color: sage-color(sage, 100);
-
-$sage-label-warning-color: sage-color(yellow, 500);
-$sage-label-warning-background: sage-color(yellow, 200);
-$sage-label-warning-border-color: sage-color(yellow, 200);
-
-$sage-label-danger-color: sage-color(red, 400);
-$sage-label-danger-background: sage-color(red, 200);
-$sage-label-danger-border-color: sage-color(red, 200);
-
-$sage-label-subtle-background: sage-color(white);
-$sage-label-subtle-draft-border-color: sage-color(grey, 400);
-$sage-label-subtle-success-border-color: sage-color(sage, 300);
-$sage-label-subtle-warning-border-color: sage-color(yellow, 300);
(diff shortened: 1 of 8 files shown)


app/assets/stylesheets/sage/system/patterns/elements/_label.scss
```diff
@@ -0,0 +1,62 @@
...
+.sage-label {
+  padding: $sage-label-padding;
+  font-size: $sage-label-font-size;
+  font-weight: $sage-label-font-weight;
+  text-align: center;
+  white-space: nowrap;
+  border-radius: $sage-label-border-radius;
+  border-width: rem(1px);
+  border-style: solid;
+  &--draft {
```
philschanely: How would you feel about moving away from this kind of shorthand with the & used for chunks of a name? I'd prefer to only use & for a complete piece of a selector but not a chunk of the name. While it is certainly elegant from a composition standpoint I find this harder to search and a little too abstract.

app/assets/stylesheets/sage/system/patterns/elements/_label.scss
```diff
@@ -0,0 +1,62 @@
...
+    color: $sage-label-warning-text-color;
+    background-color: $sage-label-warning-color;
+    border-color: $sage-label-warning-color;
+  }
+  &--danger {
+    color: $sage-label-danger-text-color;
+    background-color: $sage-label-danger-color;
+    border-color: $sage-label-danger-color;
+  }
+  &--subtle {
```
philschanely: If my first comment above is tennable what would you think about reorganiing these so that each modification when also `--subtle` is still grouped with its state like this?

```scss
.sage-label--draft {
  ...
  &.sage-label--subtle {
    ...
  }
}
```

Its not a hill I'd die on and a bit hair-splitting in this particular case since we're talking about styles on the same physical thing. But the hope is that the styles for any given element, block, or modifier are grouped together especially when context modifies them so that the settings for that thing can all be seen together rather than in several places throughout the style sheet system.

app/assets/stylesheets/sage/system/patterns/elements/_label.scss
```diff
@@ -0,0 +1,62 @@
+/* ==================================================
+  ** label
+  type: element
+
+
+================================================== */
+
+$-sage-label-subtle-background: sage-color(white);
```
philschanely, line 8: Love your variable usage! 🙌

philschanely (commented):
Some notes here to discuss. :) Looking forward to your thoughts.

app/assets/stylesheets/sage/system/patterns/elements/_label.scss
pixelflips: > How would you feel about moving away from this kind of shorthand with the & used for chunks of a name? I'd prefer to only use & for a complete piece of a selector but not a chunk of the name. While it is certainly elegant from a composition standpoint I find this harder to search and a little too abstract.

@philschanely - would you mind elaborating here just a bit? Not sure I totally follow. Thanks in advance!

app/assets/stylesheets/sage/system/patterns/elements/_label.scss
pixelflips: @philschanely - makes total sense! Thanks and I will adjust.

philschanely (approved):
Beautiful!! 🎉

PR #99, opened 2020-03-31
Card implementation

pixelflips:
Match design and functionality of card components as seen in Figma spec:

cards


app/assets/stylesheets/sage/system/patterns/objects/_card.scss
```diff
@@ -0,0 +1,98 @@
...
+  type: object
+
+
+================================================== */
+
+.sage-card {
+  margin: 0 0 rem(32px) 0;
+  border-radius: $sage-card-radius;
+  box-shadow: $sage-card-shadow;
+  &__title {
```
philschanely: Can these be removed from nesting with `&` to their own selectors like we did with modifiers on a previous PR?

app/assets/stylesheets/sage/system/patterns/objects/_card.scss
```diff
@@ -0,0 +1,98 @@
...
+    word-wrap: break-word;
+  }
+  .sage-card-body {
+    padding: $sage-card-spacing;
+    :last-child {
+      margin: 0;
+    }
+  }
+}
+.sage-card-img {
```
philschanely: See BEM comments in the markup below.

app/assets/stylesheets/sage/system/patterns/objects/_card.scss
```diff
@@ -0,0 +1,98 @@
...
+  &__title {
+    margin: 0 0 rem(16px) 0;
+    word-wrap: break-word;
+  }
+  &__text {
+    margin-bottom: rem(16px);
+    color: $sage-card-text-color;
+    word-wrap: break-word;
+  }
+  .sage-card-body {
```
philschanely: See BEM comments in the markup below.

philschanely: Also this seems to be another place where nesting should be switched so that elsewhere you have (if this nesting specificity is even needed; perhaps these are actually the default styles for the `.sage-card__body`):
```scss
.sage-card-body {
  ...

  .sage-card & {
    ...
  }
}
```

app/assets/stylesheets/sage/system/patterns/objects/_card.scss
```diff
@@ -0,0 +1,98 @@
...
+  width: 100%;
+  border-radius: $sage-card-radius $sage-card-radius 0 0;
+}
+
+// compact card
+.sage-card--compact {
+  display: flex;
+  flex: none;
+  flex-direction: row;
+  .sage-card-body {
```
philschanely: More nesting to switch if applicable.

app/views/sage/examples/objects/card/_preview.html.erb
```diff
@@ -0,0 +1,98 @@
+<div class="sage-row">
+  <div class="sage-col--md-6">
+    <div class="sage-card">
+      <img class="sage-card-img" src="/assets/sage/card/card-placeholder-lg.png" alt="">
+      <div class="sage-card-body">
```
philschanely: Any reason this couldn't be `sage-card__body` and `sage-card__img` above?

app/views/sage/examples/objects/card/_preview.html.erb
```diff
@@ -0,0 +1,98 @@
...
+        <h2 class="sage-card__title t-sage-heading-5">Card w/ Image</h2>
```
philschanely, line 6: Do you think we'll always use `t-heading-5` for this element? If so, perhaps we switch and use `@extend %t-sage-heading-5` in the stylesheet instead. Otherwise, I could see this being potentially a case where separating the two is helpful if the card title could in some places use a different type spec. 🤔

philschanely (commented):
Great start here! A few comments related to BEM and nesting style rules.

app/views/sage/examples/objects/card/_preview.html.erb
pixelflips: @philschanely - I think right now it's not entirely clear, but from my experience, I am sure the cards will evolve over time and I could seem them containing all sorts of elements and including different sizes. I think it's best to separate the 2 for now and possibly revisit if the title size is eventually set in stone.

philschanely (approved):
🏆 🏆 🏆

PR #137, opened 2020-04-02
new description element

cmcfadzean:
A very simple component that is used a lot in the super admin. 
This does not need to be the end design but looking to experiment a bit in super admin to see what issues pop up with sage and existing styles and this would be an easy one to start using.

### DIFF
--- a/app/assets/stylesheets/sage/system/patterns/elements/_description.scss
+++ b/app/assets/stylesheets/sage/system/patterns/elements/_description.scss
@@ -1,0 +1,21 @@
+/* ==================================================
+  ** description
+  type: element
+
+
+
+================================================== */
+
+.sage-description {
+  margin-bottom: sage-spacing();
+  &__title {
+    color: sage-color(charcoal, 300);
+    text-transform: uppercase;
+  }
+  &:last-of-type {
+    margin-bottom: 0;
+  }
+  @media (max-width: sage-breakpoint()) {
+    /* description responsive styles */
+  }
+}
--- a/app/assets/stylesheets/sage/system/patterns/elements/_description.scss
+++ b/app/assets/stylesheets/sage/system/patterns/elements/_description.scss
@@ -8,14 +8,17 @@
 
 .sage-description {
   margin-bottom: sage-spacing();
-  &__title {
-    color: sage-color(charcoal, 300);
-    text-transform: uppercase;
-  }
   &:last-of-type {
     margin-bottom: 0;
   }
-  @media (max-width: sage-breakpoint()) {
-    /* description responsive styles */
-  }
+}
+
+.sage-description__title {
+  @extend %t-sage-body-semi;
+  color: sage-color(charcoal, 300);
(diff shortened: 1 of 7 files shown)


app/assets/stylesheets/sage/system/patterns/elements/_description.scss
```diff
@@ -0,0 +1,21 @@
...
+  ** description
+  type: element
+
+
+
+================================================== */
+
+.sage-description {
+  margin-bottom: sage-spacing();
+  &__title {
```
philschanely: So far our team consensus is leaning away from nesting in this fashion. Consider pulling this out into a separate selector:

```scss
.sage-description__title { ... }
```

app/assets/stylesheets/sage/system/patterns/elements/_description.scss
```diff
@@ -0,0 +1,21 @@
...
+    color: sage-color(charcoal, 300);
+    text-transform: uppercase;
+  }
+  &:last-of-type {
```
philschanely, line 11: FYI Team consensus is OK with this use of `&`.

app/assets/stylesheets/sage/system/patterns/elements/_description.scss
```diff
@@ -0,0 +1,21 @@
...
+    margin-bottom: 0;
+  }
+  @media (max-width: sage-breakpoint()) {
```
philschanely: Should we remove this if its not currently used?

philschanely (commented):
Looking good. Some suggestions here for aligning with how other component styles are being written. Also on the `dt` consider adding `@extend %t-sage-body-semi;` and on `dd` consider adding `@extend %t-sage-body` just to ensure the desired styles are applied consistently and scoped :)

app/assets/stylesheets/sage/system/patterns/elements/_description.scss
cmcfadzean: Interesting. Is the reason for that for readability?

philschanely: My read on the consensus is that this makes readability and searchability more challenging for the sake of a developer shorthand. Instead we can save `&` for representing whole things rather than as a shorthand a portion of the name of a thing. For example, these would be totally fine...

```scss
.sage-description__title {
  ...
  
  &:hover {
    ...
  }

  .sage-description--danger & {
    ...
  }
}
```

...since here the `&` is alway representing the whole piece of the selector rather than it mixing where sometimes it is part of a name and other times a whole selector. 

I understand this is a change from a style that has been in use in `kajabi-products` and here in Sage too, but we're hoping to bring some consistency at least in Sage.

cmcfadzean: Sounds good to me. If you guys don't mind the extra burden of typing out the whole name then I wont fight ya on it

philschanely (approved):
🤝

PR #140, opened 2020-04-02
Sage SearchSelect

### LINKED ISSUE #85: Dropdowns
Dropdowns should be styled to match Figma spec.

Note that the autocomplete dropdown variant may not be included at this time.

Additional research:
- https://24ways.org/2019/making-a-better-custom-select-element/ (using datalists)
- https://www.24a11y.com/2019/select-your-poison/

dropdowns

AndrwM:
Resolves #85

# _Sage Dropdown! 🎉_  

- Supports html within the options (so that we can have heading/description/icons).
- a11y compatible (thanks @monkeypox8 for sharing research on this in the issue)
- Completely keyboard controllable: can be tabbed into, search is auto focused and arrow keys can be used to select options while simultaneously searching.
- Can be disabled
- Autoloads initial data

----
Here's a tiny gif, github won't allow any larger filesizes 👅 

screencast 2020-04-24 10-14-42


AndrwM:
Pausing on this to focus on 🕷**Build Bug Brigade** 🕷

app/views/sage/examples/elements/dropdown/_preview.html.erb
```diff
@@ -0,0 +1,79 @@
...
+      <div class="sage-dropdown__search-container">
+        <input class="sage-dropdown__search-field"
+          type="text"
+          placeholder="Search..."
+          data-js-sagedropdown-search
+        >
+      </div>
+      <ul class="sage-dropdown__option-container"  data-js-sagedropdown-optioncontainer>
+        <li class="sage-dropdown__option" data-js-sagedropdown-option="">
+          <div class="t-sage-body" style="opacity: 0.4">(Empty)</div>
```
AndrwM, line 20: I wasn't completely sure how to handle this...Thinking I need some type of utility to style "greyed out" text. Waiting for feedback from design on the final call on how we're handling an empty value as it's not included in the design spec.

philschanely (approved):
Beautiful and clean implementation here! 🙌 

One thought I had: this is just fine as-is. What if either instead of this, or perhaps as another option for implementation the source starts as a `select` and JS transforms to this, or at least generates this, hides the original, but keeps changes in sync? Just thinking about a) progressive enhancement such that the select is the default and fully functional without JS (albeit bland) and b) generating a selection with options is a neat initial way of writing a dropdown for use in forms. 🤔 

Regardless, this is 👍 as far as I'm concerned!

AndrwM:
@philschanely That's a really great idea. It would also reduce the need for custom-generated markup in the form partial.

One goal that I had in mind when working on this project is the expectation that these dropdown options could have icons & descriptions because the content could vary from just being a string. I do wonder if there's a middle-ground here where we could add that markup within an html attribute of each `<option>`.

Thanks for mentioning this. I'll keep this on my radar as we begin implementing this in simpleform.

philschanely:
@AndrwM Just remembered that all variables will need to be `var` rather than `const` or `let` as we move over to `products`.

AndrwM:
@philschanely I had a feeling this might come up. 🙂

3/4 of the way through this PR I converted it to ES6 syntax because the amount of scope binding I was doing was becoming very verbose. (Mostly a lot of conversion to [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions).) I was able to reduce the lines of the JS by about 1/3 and it made the document much more readable.

Since we're expecting Webpacker support soon and we don't need to support anything except modern browsers (JS-wise) within super_admin, my thinking is that it's better to take a more modern approach in the name of maintainability/readability.

At the moment there's solid ES6 support out-of-the-box for all browsers except IE11:
https://kangax.github.io/compat-table/es6/

AndrwM:
I'm planning to hold off merging until everyone can weigh in here.

pixelflips:
@AndrwM - this is great! I am wondering though, in regards to the ES6, I think it's less about browser support and more about the `kajabi-products` build failing when we use ES6 syntax (I previously broke the `kajabi-products` build just by using `let`). Do you know how soon Webpacker support might happen? I'm just wondering if we will need to hold on merging this until that happens?

AndrwM:
@pixelflips Hmmm.. I see. 😦 
I'm definitely open to putting this on hold until we can land webpacker support. I would prefer that to shipping less tidy version of this, personally; this format is ultimately the world we want to live in.

I've been doing some reading/thinking about this since my initial sprint and I'm feeling hopeful that this could come soon.

AndrwM:
Additionally, I think a Webpacker implementation would allow us to use the new SCSS (`@use`, local var support, etc)..come to think of it.

PR #139, opened 2020-04-03
Revise input field and textarea label appearance

teenwolfblitzer:
## Description
Following discussions with design, the focused/floating label styles will be tweaked to resemble the Material Design "outlined" appearance, as can be seen here: https://material.io/components/text-fields/

In addition, input fields will be reduced in height to better align when displayed inline alongside buttons (corrected in #143).

Both text input fields and text areas will be updated to match.

Revised design:
text-fields


philschanely (approved):
🙌

PR #143, opened 2020-04-03
Button outline inconsistencies

teenwolfblitzer:
Button outline height is larger than Figma spec, which calls for a `40px` default height, and a `50px` height (including outline) when focused. Current default state height is `44px` with a total focused height of `57px` with outline.

This will be more noticeable once the revised input field heights (#139) are merged.

Note that heights displayed in figma do not include borders, so here `48px` + `1px` border on each side = `50px`.
button-size-figma


pixelflips (approved):
👍 LGTM! Nice catch too btw!

PR #144, opened 2020-04-06
Live stream video grid

### LINKED ISSUE #142: Add Live Stream Video and Grid
Add layout configurations for videos in the Live application.

[Design spec](https://www.figma.com/file/ikwcMcE6HEb653b0oVCgy6/%E2%9A%99%EF%B8%8FKajabi-Live-SAGE?node-id=89%3A363)

Screen Shot 2020-04-03 at 3 20 46 PM

philschanely:
Closes #142 

Screen Shot 2020-04-06 at 3 20 27 PM


app/assets/stylesheets/sage/system/patterns/objects/_live_stream_video_grid.scss
```diff
@@ -0,0 +1,90 @@
...
+  @include position-full;
+}
+
+.sage-live-stream-video-grid__tile {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  overflow: hidden;
+  position: relative;
+  border-radius: rem(4px);
```
pixelflips: Not a show stopper for sure, but I am curious if since now we have the rem function, would that be the preferred method for setting a radius vs. using the radius token?

philschanely: Good catch! I'll switch this

app/assets/stylesheets/sage/system/patterns/objects/_live_stream_video_grid.scss
```diff
@@ -0,0 +1,89 @@
...
+  
+  @for $i from 1 to 6 {
```
teenwolfblitzer: I think you want `1 through 6` rather than `to` here; the 6th iteration is not being generated.

app/assets/stylesheets/sage/system/patterns/objects/_live_stream_video_grid.scss
```diff
@@ -0,0 +1,89 @@
...
+  @include position-full;
+}
+
+.sage-live-stream-video-grid__tile {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  overflow: hidden;
+  position: relative;
+  border-radius: rem(4px);
```
teenwolfblitzer: Can use `sage-border(radius)` here.

teenwolfblitzer (commented):
🚀 Looking great! One suggestion and one item to fix on the for loop for the 6th item in the grid, should be good to go once that's in.

pixelflips (approved):
👍 LGTM - Nice work!

app/assets/stylesheets/sage/system/patterns/objects/_live_stream_video_grid.scss
teenwolfblitzer: Looks like @pixelflips already caught this one!

app/assets/stylesheets/sage/system/patterns/objects/_live_stream_video_grid.scss
philschanely: OOOO good catch! Fixed.

teenwolfblitzer (approved):
👍 🔥

PR #129, opened 2020-04-06
Refactor sidebar JS as global module

teenwolfblitzer:
## Description
This originally started as a refactor to move the sidebar/side menu JS from the Sage documentation, but ended up becoming a larger restructure for our JS to play nicely with our current asset pipeline.

It may all be moot once we (eventually) move to webpack, but for now this will allow us to conditionally load JS modules without the use of ES6 imports and/or a new build process.

- Restructure JS to use an adapted revealing module pattern
  - JS namespacing using `Sage` global
  - modules conditionally initialized in router function
  - functions in modules are private (not scoped to window) by default, but can selectively expose public functions for consumption (ex: type `Sage` in your browser console to see what's visible)
  - added `util.js` for global utility functions
- Fix error in console from `.sage-live-option-menu-anchor` event listener misfire

FUTURE/discussion:
- Probably should default to using strict mode
- Named functions for event listeners (vs anonymous functions)
- Comment/documentation patterns - model after Sass?


app/assets/javascripts/sage/system/define.js
```diff
@@ -0,0 +1,2 @@
+var Sage = Sage || {};
```
AndrwM, line 1: 😍 😍

app/assets/javascripts/sage/system/util.js
```diff
@@ -0,0 +1,13 @@
...
+  function getBtnTarget(e) {
+    return e.target.getAttribute('aria-controls') ? e.target.getAttribute('aria-controls') : e.target.getAttribute('data-js-menu');
+  }
+
+
+  return {
+    getBtnTarget: getBtnTarget
+  };
+
+})(Sage);
```
AndrwM: Love having a place to hold utility funcs. There are several bits incoming from the Dropdown PR that would be useful to make globally available.

app/assets/stylesheets/sage/system/patterns/objects/_sidebar.scss
```diff
@@ -4,53 +4,45 @@
...
+  transition: transform 0.2s ease-out, opacity 0.15s ease-in-out;
+  opacity: 0;
 
-  &--md {
-    @extend .sage-sidebar;
-    width: sage-sidebar(md);
+  &.is-open {
+    transform: none;
+    opacity: 1;
   }
```
AndrwM: Should this be `.sage-sidebar--open`?

AndrwM (approved):
Awesome! Loving the JS organization.

app/assets/stylesheets/sage/system/patterns/objects/_sidebar.scss
teenwolfblitzer: Oh, great question. From past experience, with state classes I'm of the opinion we should stray from the BEM model, and go with SMACSS by using `.is-` and `.has-` naming. The argument being that the modifier is user-generated from interaction rather than as a style modification being applied to the default/normal state. But more importantly, I feel the naming is more clear and concise—names for BEM state classes tend to get lengthy, and can also be confused with styling modifier classes

/cc @philschanely

AndrwM: I like the idea of having a naming strategy for interactivity-related states if the rest of the team feels good about it also.

I do worry that we may run into leakage styling child BEM nodes if we want to tie them to the parent BEM node interactivity state.
```
.sage-hotdog__bun {
  ...

  .has-relish & {
     ...
  }
}
```
In the instance above `.has_relish` isn't namespaced to a BEM object/element so there could be a collision from a parent node. We'll have to be sure to watch out for this type of leak.

teenwolfblitzer: 🌭Agreed—we'd want to keep these state classes confined within individual components to avoid collision. It's a good question for the group, curious to see what others think.

app/assets/javascripts/sage/docs/live-option-menu.js
```diff
@@ -0,0 +1,27 @@
+Sage.docs.liveOptionMenu = (function() {
```
philschanely, line 1: 😍

philschanely (approved):
Stellar work! 🌟

app/assets/stylesheets/sage/system/patterns/objects/_sidebar.scss
philschanely: @monkeypox8 and @AndrwM I have never actually understood exactly why we'd do `is-open` instead of as a modifier `--open` other than distinguishing a purely JS toggle. I see the "state" argument as a part of the SMACSS convention, but do feel it has a conflict with BEM in this. I'd propose that if the state is something broader than the component itself the `is-open` is OK, but if its really a state of the thing in particular, and its having an affect on the CSS styles I'd advocate for the modifier instead. Def open to what you all think best!

cmcfadzean: For what its worth I tend to agree with @philschanely on this one but I'm not terribly opposed to either option

teenwolfblitzer: Appreciate all the input! Sounds like the consensus is to stick with BEM using modifiers for state, so I'll change this to match. I do feel that it can be helpful to stray from BEM when beneficial, but as @AndrwM pointed out, there's a pretty substantial drawback for class collision, especially once Sage is released outside of our circle.

philschanely: @monkeypox8 Absolutely. Key to me is that we're on the same page about when and why we stray from BEM. Such as with our type classes. I'd propose this rubric here:

* Is the class JS-only? Use state form or a `data-js` setting
* If not (CSS changes apply), is it scoped or controlled broader than this component? Use state form.
* Otherwise, use BEM modifier.

cmcfadzean (approved):
Excited for this

PR #154, opened 2020-04-09
Quick Fix: Duplicate max-widths for containers.

pixelflips:
Update` _container.scss` to remove duplicate max-widths from being applied to the different `.sage-container` sizes.

Screen Shot 2020-04-09 at 11 40 51 AM


teenwolfblitzer (approved):
👍 Great work and nice find!

PR #127, opened 2020-04-10
[mobile] Tables overflowing container

teenwolfblitzer:
## Description
- Fixes status table overflowing page container and breaking layout on iOS
- Fixes status table key overflow
- Adds base styles for tables, including row striping and responsive options

### Screenshots
|  Before   |  After  |
|--------|--------|
ios-table-scroll|table-fix|

Adapted styling from Figma playground
figma-table

### DIFF
--- a/app/assets/stylesheets/sage/system/patterns/elements/_table.scss
+++ b/app/assets/stylesheets/sage/system/patterns/elements/_table.scss
@@ -7,21 +7,25 @@
 .sage-table {
   --table-cell-padding: #{sage-spacing(sm)} #{sage-spacing(xs)};
 
-  table-layout: initial;
   width: 100%;
   max-width: $sage-table-max-width;
 
+  caption {
+    @extend #{$sage-table-caption-font-size};
+    text-align: $sage-table-caption-alignment;
+  }
+
   thead {
-    color: $sage-table-heading-text-color;
+    color: $sage-table-heading-font-color;
     border-bottom: $sage-table-border;
 
     td {
-      @extend %t-sage-heading-6;
+      @extend #{$sage-table-heading-font-size};
     }
   }
 
   tbody {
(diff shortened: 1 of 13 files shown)


app/assets/stylesheets/sage/system/patterns/elements/_table.scss
```diff
@@ -0,0 +1,86 @@
+/* ==================================================
+  ** sage_table
+  type: element
+
+================================================== */
+
+.sage-table {
+  --table-cell-padding: #{sage-spacing(sm)} #{sage-spacing(xs)};
+
+  table-layout: initial;
```
monkeypox8, line 10: This reset may not be necessary

app/views/sage/examples/elements/table/_preview.html.erb
```diff
@@ -0,0 +1,89 @@
+<h3 class="t-sage-heading-6">Responsive table</h3>
+<div class="sage-table--responsive">
```
philschanely: @monkeypox8 I see that the double-nesting is necessary, but the BEM classes here don't seem quite right to me. I wonder if we could switch to something like `sage-table-wrapper` as the outer and `sage-table-wrapper__overflow-frame` or something like that for the inner? It just seems like not BEM to have outer elements be an element and a modifier of the inner block. 🤔

philschanely (changes requested):
Beautiful responsive work but I think the classnames should be adjusted. See comment.

app/views/sage/examples/elements/table/_preview.html.erb
teenwolfblitzer: Oh, excellent point. I was going back and forth on naming earlier on, so the way I left it feels odd. I'll make the change.

philschanely (approved):
🙌

PR #158, opened 2020-04-10
Revise tables in documentation to use `.sage-table` styling

teenwolfblitzer:
## Description
- Revised documentation and generators to use new table styling
- Addresses accessibility of status table labels
- Slight refactor of table CSS and variable cleanup

### Screenshots
|  Before   |  After  |
|--------|--------|
|table-old|table-new|

## Related issues
- #127

### DIFF
--- a/app/assets/stylesheets/sage/system/patterns/elements/_table.scss
+++ b/app/assets/stylesheets/sage/system/patterns/elements/_table.scss
@@ -5,7 +5,8 @@
 ================================================== */
 
 .sage-table {
-  --table-cell-padding: #{sage-spacing(sm)} #{sage-spacing(xs)};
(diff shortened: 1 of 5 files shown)


pixelflips (approved):
👍 LGTM - Nice Work!

app/assets/stylesheets/sage/system/patterns/elements/_table.scss
```diff
@@ -15,12 +16,25 @@
     text-align: $sage-table-caption-alignment;
   }
 
-  thead {
+  thead,
```
teenwolfblitzer, line 19: May need to revisit using classes over elements here once we have direction from design on table styles. It's also possible we may not have full control over adding classes to `thead` and `tbody` elements with the existing codebase.

PR #163, opened 2020-04-13
Sage.init()

AndrwM:
### What this includes:
- DRY up Rails layouts (including partial includes for scripts & styles)
- Require explicit initialization for each Sage element script
  - This allows us to control script initialization individually on the `kajabi-products` side in the event a script:
    - is causing errors
    - is still in development
    - a smaller selection is needed for a specific `kajabi-products` view
    - a script is only required for the documentation side

### DIFF
--- a/app/assets/javascripts/sage/system/init.js
+++ b/app/assets/javascripts/sage/system/init.js
@@ -1,0 +1,20 @@
+Sage.init = function(elementNamesToInit) {
+  var shouldInit = function(elementName, selector) {
+    return elementNamesToInit.contains(elementName) && document.querySelector(selector) !== null;
+  };
+
+  // Initialize Tooltip
+  if ( shouldInit('tooltip', '[data-tooltip]') ) {
+    Sage.tooltip();
+  }
+
+  // Initialize Sidebar
+  if ( shouldInit('sidebar', '.sage-sidebar') ) {
+    Sage.sidebar.init();
+  }
+
+  // Initialize Overlay
+  if ( shouldInit('overlay', '.sage-overlay') ) {
+    Sage.overlay.init();
+  }
+}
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/sage/system/init.js
```diff
@@ -0,0 +1,20 @@
+Sage.init = function(elementNamesToInit) {
+  var shouldInit = function(elementName, selector) {
```
teenwolfblitzer, line 2: 😍

app/views/layouts/sage/_scripts.html.erb
```diff
@@ -0,0 +1,14 @@
+<script src="//cdn.jsdelivr.net/gh/google/code-prettify@master/loader/run_prettify.js"></script>
```
teenwolfblitzer: What are your thoughts on bundling this into the `sage_docs` JS locally rather than including the external dependency? Probably a negligible impact considering the CDN's speed, but it'd be one fewer DNS request.

philschanely (approved):
🔥 🔥 🔥

teenwolfblitzer (approved):
👍 🔥 Love this. One non-blocking item to ponder with bundling the `run_prettify.js` script. Nice work!

cmcfadzean (commented):
My only thought is that the styles and js partials should be in the `app/views/sage/application` folder

AndrwM:
@cmcfadzean Thanks! Updating

app/views/layouts/sage/_scripts.html.erb
AndrwM: Agreed, this makes the most sense to be a npm dependancy. I'm going to push this to another PR though – we're production dependency free at the moment and I'm thinking there may need to be some coordination with SRE to ensure our gem's npm dependancies load into kajabi-product smoothly.

PR #169, opened 2020-04-14
Grid alignment classes

teenwolfblitzer:
## Description
Adds options for alignment and column visibility for feature parity with current site. Probably as far as we can take this without going full utility classes.

- [x] classes for vertical alignment of rows: top (default), middle, bottom
- [ ] ~classes for horizontal alignment of children: left (default), center, right~
- [x] ability to show and hide columns per breakpoint
- [x] alignment uses logical properties with fallback when possible

## Related issues
- Full documentation for grid use and features to be added in #168

### DIFF
--- a/app/assets/stylesheets/sage/system/layout/_grid.scss
+++ b/app/assets/stylesheets/sage/system/layout/_grid.scss
@@ -12,8 +12,8 @@
   display: flex;
   flex-direction: row;
   flex-wrap: wrap;
-  margin-left: calc(#{$sage-grid-gap-md} / -2);
-  margin-right: calc(#{$sage-grid-gap-md} / -2);
+  margin-left: calc(#{$sage-grid-gap} / -2);
+  margin-right: calc(#{$sage-grid-gap} / -2);
 }
 
 .sage-row {
@@ -50,7 +50,7 @@
 @mixin sage-col {
   flex: 1 1 auto;
   max-width: 100%;
-  padding: 0 calc(#{$sage-grid-gap-md} / 2);
+  padding: 0 calc(#{$sage-grid-gap} / 2);
 }
 
 @for $sage-i from 1 through $sage-grid-columns {
@@ -77,7 +77,7 @@
   text-align: end;
 }
 
-@media (min-width: sage-breakpoint(sm-min)) {
+@media (min-width: $sage-grid-breakpoint-sm) {
   @for $sage-i from 1 through $sage-grid-columns {
     .sage-col--sm-#{$sage-i} {
       flex: 0 1 auto;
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/sage/system/layout/_grid.scss
```diff
@@ -12,34 +12,72 @@
...
 }
 
-@media (min-width: sage-breakpoint(sm-min)) {
+.sage-col--align-start {
+  text-align: left;
+  text-align: start;
+}
+.sage-col--align-end {
+  text-align: right;
+  text-align: end;
```
pixelflips: any reason for two `text-align` here?

app/assets/stylesheets/sage/system/layout/_grid.scss
```diff
@@ -12,34 +12,72 @@
...
+  .sage-col-#{$sage-i} {
+    flex: 0 1 auto;
+    width: percentage($sage-i / $sage-grid-columns);
+  }
 }
 
-@media (min-width: sage-breakpoint(sm-min)) {
+.sage-col--align-start {
+  text-align: left;
+  text-align: start;
```
pixelflips: also two `text-align` here?

philschanely: Good call @pixelflips . @monkeypox8 should this be scoped in the docs sheet instead?

philschanely (approved):
Pending the text-alignment question, this looks good to me!

app/assets/stylesheets/sage/system/layout/_grid.scss
teenwolfblitzer: Ah, great question. Since the second declaration uses logical property syntax, which older browsers (IE11 and some legacy Edge) don't/won't support, they'll resort to `text-align: right`. Normally we could leave out IE11 support as long as the core is at least functional, but layout alignment feels significant enough to have at least a fallback.

pixelflips: @monkeypox8 - thanks for clarifying!

pixelflips (approved):
👍 +1

app/assets/stylesheets/sage/system/layout/_grid.scss
teenwolfblitzer: Now that I think about it, it may make sense to create some mixins to accomplish this; we should be able to `@supports` the logical props and output what the browser can understand.

app/assets/stylesheets/sage/system/layout/_grid.scss
philschanely: @monkeypox8 I suppose I still need clarification on why we'd bake text-alignment into the grid system. Should that be either broader in a reset file or, as needed, more narrowly in a component?

teenwolfblitzer: @philschanely Hmm, that's true… this may be more opinionated than needed. The way I was approaching it was to duplicate pieces of functionality in the existing grid, which (using Bootstrap) allows for horizontal/vertical alignment at the container level. But in our case, most of that should be accomplished in a component rather than in the underlying structure. I feel like there should be _some_ options available for setting horizontal alignment, but adding those elsewhere feels even more like we'd be heading down the path of utility classes.

For now, I'll take out the horizontal formatting until we have a chance to decide as a group whether we even need to cross this bridge.

philschanely: I'm OK if we keep for now but revisit at a later date to consider moving such things lower down. For example, we now have the `.sage-type` container for open text. We could consider moving text-alignment here such as:
```scss
.sage-type {
  ...
  .sage-col--align-end & {
    ...
  }
}
```

teenwolfblitzer: Going to remove for now and revisit down the road. Text alignment at this level feels like it could be too far-reaching if left in to cascade.

PR #173, opened 2020-04-16
Avatar Updates

philschanely:
- Updates Avatars to better comply with current needs (see comments inline)
- Minor patches to other aspects for application in Live


app/assets/stylesheets/sage/system/patterns/objects/_live_avatar.scss
```diff
@@ -35,6 +35,11 @@ $-live-avatar-ring-colors: (
 );
 $-live-avatar-tile-size: rem(216px);
 
+@mixin live-avatar-scaling() {
```
philschanely, line 38: Moved the scaling settings to a mixin to preserve and isolate for future use when specific scale sizes are implemented.

app/views/sage/examples/objects/live_avatar/_preview.html.erb
```diff
@@ -58,70 +58,4 @@
     scale: false,
     custom_class: ""
   %>
-</div>
```
philschanely, line 61: Removed the scaling demonstrations; will return to this when we have particular sizes to display. Scaling as needed is now only baked into the `--tile` modifier.

app/assets/stylesheets/sage/system/vendor/_reboot.scss
```diff
@@ -266,6 +266,7 @@ button::-moz-focus-inner,
 input[type="radio"],
 input[type="checkbox"] {
   box-sizing: border-box;
+  margin: 0;
```
philschanely, line 269: Similar level of specifity sets a top margin in current products styles at this time. Overriding.

teenwolfblitzer: Interesting. I see the temporary shim here: https://github.com/Kajabi/kajabi-products/blob/7b54cd086ae8fc491f9021d873854c0ef0a7ea01/.storybook/preview-head.html#L5. Is the margin affecting only the Sage inputs? If so, I would have hoped that the margin reset in `.sage-checkbox__input` and `.sage-radio__input` would suffice.

pixelflips (approved):
👍 LGTM!

teenwolfblitzer (approved):
👍 LGTM, just a question about the margin reset in the reboot.

app/assets/stylesheets/sage/system/vendor/_reboot.scss
philschanely: @monkeypox8 sadly the specificity in `products` now is `0 1 1` since what's there is `input[type...]`. So rather than increasing our BEM element's specificity artificially I thought I'd just cut it off at the reboot level. I'll merge for now but happy to move forward with an improvement later if you'd like this a different way.

PR #175, opened 2020-04-16
Jcm pagination

cmcfadzean:
Screen Shot 2020-04-16 at 2 20 02 PM


app/assets/stylesheets/sage/system/patterns/objects/_pagination.scss
```diff
@@ -0,0 +1,46 @@
...
+}
+
+.sage-pagination__link--disabled {
+  color: sage-color(grey, 400);
+}
+
+.sage-pagination__icon {
+  display: inline-block;
+  transform: translateY(2px);
+}
```
AndrwM: It might be best here if we used `flex` alignment instead of hardcoding the offset in the event that the body font size changes (via a mobile mediaquery or changing design).

AndrwM (approved):
Left a comment about 1 issue but otherwise looks great. 🤩 +1

pixelflips (approved):
👍 +1 Looks pretty clean to me. Nice work!

app/assets/stylesheets/sage/system/patterns/objects/_pagination.scss
```diff
@@ -0,0 +1,48 @@
...
+  padding: 0 sage-spacing(xs);
+}
+
+.sage-pagination__link--disabled {
+  color: sage-color(grey, 400);
+}
+
+.sage-pagination__icon {
+  display: inherit;  
+  padding: 0 sage-spacing(2xs)
```
teenwolfblitzer: Need a semicolon here

app/views/sage/examples/objects/pagination/_preview.html.erb
```diff
@@ -0,0 +1,32 @@
...
+      <a href="#" class="sage-pagination__page" aria-label="Go to Page 1">1</a>
+    </li>
+    <li>
+      <a href="#" class="sage-pagination__page" aria-label="Go to Page 2">2</a>
+    </li>
+    <li>
+      <a href="#" class="sage-pagination__page" aria-label="Go to Page 3">3</a>
+    </li>
+    <li>
+      <span class="sage-pagination__page sage-pagination__page--gap" aria-current="true">...</span>
```
teenwolfblitzer: Example here is setting `aria-current="true"` on page 20 as well, was it intended to also be on this item?

app/views/sage/examples/objects/pagination/_preview.html.erb
```diff
@@ -0,0 +1,32 @@
+<nav class="sage-pagination" role="navigation" aria-label="Pagination Navigation">
```
teenwolfblitzer: We can remove "Navigation" from the `aria-label` and the `role`. The role is unnecessary with the element, and a screenreader will read the element and the label as a duplicate, so it would read "Pagination Navigation, Navigation"

app/views/sage/examples/objects/pagination/_preview.html.erb
```diff
@@ -0,0 +1,32 @@
...
+      <span class="sage-pagination__page sage-pagination__page--gap" aria-current="true">...</span>
+    </li>
+    <li>
+      <a href="#" class="sage-pagination__page" aria-label="Go to Page 18">18</a>
+    </li>
+    <li>
+      <a href="#" class="sage-pagination__page" aria-label="Go to Page 19">19</a>
+    </li>
+    <li>
+      <span class="sage-pagination__page sage-pagination__page--active" aria-current="true">20</span>
```
teenwolfblitzer: I believe `aria-current="page"` may be more fitting for this since it's specifically for pagination

app/views/sage/examples/objects/pagination/_preview.html.erb
```diff
@@ -0,0 +1,32 @@
...
+    </li>
+  </ul>
+  <span class="sage-pagination__link sage-pagination__link--disabled">
```
teenwolfblitzer: Ideally for the disabled links, we would want to still keep the same `<a>` anchor element as the others, but add in `aria-disabled="true"`. Then on the `.sage-pagination__link--disabled`, setting `pointer-events` should disable the click

teenwolfblitzer (commented):
Looks great, the missing semicolon was preventing me from spinning up the server. Other than that, couple of non-blocking items to consider that can be addressed in a follow-up.

philschanely (approved):
@monkeypox8 has made a great a11y review; otherwise 👍 from me.

PR #176, opened 2020-04-17
System banner/alert styles

teenwolfblitzer:
## Description
Creates component used for sitewide/systemwide banners, resulting from in-progress work for user masquerading.

- Renames the previous "banner" object to "alert", matching with Figma and proposed use

banners

## Related issues:
- Affects #179


app/assets/javascripts/sage/sage_docs.js
```diff
@@ -13,5 +14,8 @@ if (document.querySelector('.sage-docs') !== null) {
     Sage.docs.liveOptionMenu.init();
   }
 
+  if (document.querySelector('.sage-banner--active') !== null && document.querySelector('.example__preview--page') !== null) {
```
teenwolfblitzer, line 17: Normally this would be initialized in `init.js` like the other modules. For the purposes of the documentation site, I've limited the JS init to the preview, so that the banner doesn't pop up when viewing the full list of objects. Open to recommendations on a better way to approach this.

app/assets/javascripts/sage/system/example.js
```diff
@@ -0,0 +1,38 @@
+Sage.module = (function() {
+  /*
+    This is an example template. To create a module:
```
philschanely, line 3: 😍

app/assets/stylesheets/sage/system/patterns/objects/_banner.scss
```diff
@@ -1,53 +1,81 @@
...
-  border-top: 4px solid;
-  border-bottom-left-radius: sage-border(radius);
-  border-bottom-left-radius: sage-border(radius);
-  box-shadow: sage-shadow();
+  display: flex;
+  align-items: center;
+  height: $sage-banner-height;
+  width: 100%;
+  padding: sage-spacing(xs) sage-spacing(sm);
+  font-weight: sage-font-weight(semibold);
```
philschanely: Should we extend `t-body-semi` here perhaps also to ensure we have the expected type spec applied?

philschanely (approved):
One small change/question but 👍 otherwise.

PR #181, opened 2020-04-17
Simpleform Compatibility - Pt 1

AndrwM:
#### ✅ Satisfies dependancies for [PR #14066](https://github.com/Kajabi/kajabi-products/pull/14066)
-----

### What this includes:
This includes styles to support our initial pass at Simpleform + Sage integration PR. It's more a PR of various bugfixes and support patches than fully-fleshed out support for all Sage inputs with Simpleform. Once I get the initial concept approved and landed, I'll be opening a follow up PR on Kajabi-Products that converts many super user forms...inevitably there will be another round of compatibility adjustments coming through in Sage when that happens.

- `.t-sage-body-danger` classes for red danger text
- `position()` mixin that we discussed as a team last week.
- Simpleform error handling support for `sage-input`
- Support for messages to be appended inside of `sage-input` (simpleform generated hint & error)
- Simpleform error handling support for `sage-switch`
- Support for messages inside of `sage-switch` (simpleform generated hint & error)

### DIFF
--- a/app/assets/stylesheets/sage/system/core/_mixins.scss
+++ b/app/assets/stylesheets/sage/system/core/_mixins.scss
@@ -110,6 +110,32 @@
   left: $offset;
 }
 
+
+/* ==================================================
+** Positioning shorthand
+  Usage: `@include position-full(10px, auto, unset, auto);`
+
+  Sets an element's position with margin/padding shorthand (top, right, bottom, left)
+  arguments: $offset | [length]
+================================================== */
+
+@mixin position($top, $right, $bottom, $left) {
+  top: _position-print-value("top", $top);
+  right: _position-print-value("right", $right);
+  bottom: _position-print-value("bottom", $bottom);
+  left: _position-print-value("left", $left);
+}
+
+@function _position-print-value($key, $value) {
+  @if not $value {
+    @error 'position() mixin requires all 4 position values (top, right, bottom, left): #{$key} is missing.';
+  } @else if $value != unset {
+    @return $value;
+  } @else {
+    @return false;
+  }
+}
+
 /* ==================================================
 ** Icon base
   Usage: `@include icon-base();`
@@ -118,8 +144,8 @@
   to displaying the icon. Usually implemented on a pseudo-element.
 
   If an `$icon` is passed in this mixin also displays the icon as content for the
-  element. 
-  
+  element.
+
   arguments: $icon | null|[string]
 ================================================== */
 
--- a/app/assets/stylesheets/sage/system/core/_mixins.scss
+++ b/app/assets/stylesheets/sage/system/core/_mixins.scss
@@ -128,7 +128,7 @@
 
 @function _position-print-value($key, $value) {
   @if not $value {
-    @error 'position() mixin requires all 4 position values (top, right, bottom, left): #{$key} is missing.';
+    @error "position() mixin requires all 4 position values (top, right, bottom, left): #{$key} is missing or invalid.";
   } @else if $value != unset {
     @return $value;
   } @else {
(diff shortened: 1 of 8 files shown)


app/assets/stylesheets/sage/system/core/_mixins.scss
```diff
@@ -118,8 +144,8 @@
   to displaying the icon. Usually implemented on a pseudo-element.
 
   If an `$icon` is passed in this mixin also displays the icon as content for the
-  element. 
-  
+  element.
+
```
AndrwM, line 148: My editor automagically trimmed the extra space here 🤷

app/assets/stylesheets/sage/system/core/_typography.scss
```diff
@@ -114,6 +114,11 @@ $type-specs: (
 
   .t-sage-#{$spec-name} {
     @extend %t-sage-#{$spec-name};
+
+    &-danger {
```
philschanely: Just to keep our convention of avoiding this kind of use of `&` consistent, could we switch this to:

```scss
... {
  .t-sage-#{$spec-name} {
    @extend %t-sage-#{$spec-name};
  }
  .t-sage-#{$spec-name}-danger {
    @extend %t-sage-#{$spec-name};
    color: sage-color(red, 300);
  }
}
```

Alternatively, would it be practical to switch and use this as a modifier on relevant elements rather than on type specs? @monkeypox8 and @pixelflips any thoughts?

philschanely (approved):
Looks good! One non-blocking comment on the `-danger` addition to the type spec.

app/assets/stylesheets/sage/system/core/_typography.scss
teenwolfblitzer: Interesting… I think I'd lean towards favoring a modifier on the element itself vs adding to the type spec here, since it isn't necessarily a typographic style. I do see the benefits of having a generic class for form styling, though. In that case, it's possible a standard error/danger style should live in `_form_section` or in its own module. `.sage-form-error`?

app/assets/stylesheets/sage/system/core/_mixins.scss
```diff
@@ -110,6 +110,32 @@
...
+@mixin position($top: false, $right: false, $bottom: false, $left: false) {
+  top: _position-print-value("top", $top);
+  right: _position-print-value("right", $right);
+  bottom: _position-print-value("bottom", $bottom);
+  left: _position-print-value("left", $left);
+}
+
+@function _position-print-value($key, $value) {
+  @if not $value {
+    @error 'position() mixin requires all 4 position values (top, right, bottom, left): #{$key} is missing.';
```
teenwolfblitzer: Stylelint is being nitpicky and would like these to be in double-quotes.

We may want to make this rule a warning rather than allowing it to break the server. Open to either option.

teenwolfblitzer (commented):
Looks great so far! One small hiccup from stylelint, should be good to go afterwards.

app/assets/stylesheets/sage/system/core/_typography.scss
AndrwM: This has been addressed. Thanks @philschanely & @monkeypox8 !

app/assets/stylesheets/sage/system/core/_mixins.scss
AndrwM: I'm more partial to it causing the server to throw an exception, that way mis-implemented `position()` solutions cannot be shipped rather than a warning showing up in the console and potentially being missed.

Fixed the stylelint issue.

app/assets/stylesheets/sage/system/patterns/elements/_form_input.scss
```diff
@@ -105,6 +105,13 @@
   }
 }
 
+.sage-input__message {
+  @extend %t-sage-body-small;
+
+  .sage-input--error & {
```
philschanely, line 111: 😍

philschanely (approved):
Well done!

PR #182, opened 2020-04-19
"Skip to content" link

teenwolfblitzer:
Adds a few accessibility updates:
- Provides a skip link to bypass navigation and proceed directly to content, visible to screenreaders.
- Revises some containers to use semantic elements
- Updates assistant with `erb` includes for markup preview


philschanely (approved):
👍

PR #188, opened 2020-04-21
Jcm panel updates

cmcfadzean:
Minor updates to panels to have a header and footer and also title styles.

### DIFF
--- a/app/views/sage/examples/objects/panel/_preview.html.erb
+++ b/app/views/sage/examples/objects/panel/_preview.html.erb
@@ -1,4 +1,11 @@
 <div class="sage-panel">
-  <h3 class="sage-panel__title">Panel Title</h3>
-  <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.</p>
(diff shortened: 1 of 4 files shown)


app/views/sage/examples/objects/panel/_preview.html.erb
```diff
@@ -1,4 +1,11 @@
 <div class="sage-panel">
-  <h3 class="sage-panel__title">Panel Title</h3>
-  <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.</p>
+  <div class="sage-panel__header">
+    <h3 class="sage-panel__title">Panel Title</h3>
```
AndrwM: @philschanely mentioned something about this here:
https://github.com/Kajabi/kajabi-products/pull/14066#discussion_r413041419

I think it might be best to use the Sage Heading component as a child of  `sage-panel__header`. So that these Panels could have more hierarchical flexibility. I could see there being a view with a big panel that has a h4 heading size and underneath 2 panels in columns with h5 heading sizes.

AndrwM (commented):
Left a comment about panel titles, otherwise looks great.

app/views/sage/examples/objects/panel/_preview.html.erb
cmcfadzean: sounds good. I will remove the title classes and have panel simply be the wrappers for content. I dig it

PR #189, opened 2020-04-21
Remove default `sage-btn` margins

### LINKED ISSUE #190: Button sizing options
## Description
- Small variant of button
  - Refactor `banner.scss` to match
- Full-width buttons (mobile viewports only?) TBD

## Related issues
- #189

teenwolfblitzer:
## Description
Base `.sage-btn` styles include bottom and right margin settings that should be scoped to specific use cases. For example, in Sage documentation and within specific modules.

When applied in the super admin, these margins push the buttons out of alignment with other inline elements.

- Addresses button sizing from #190
- Updates preview markup using includes

## Related issues
- Resolves #190

### DIFF
--- a/app/assets/stylesheets/sage/system/patterns/elements/_button.scss
+++ b/app/assets/stylesheets/sage/system/patterns/elements/_button.scss
@@ -96,10 +96,9 @@ input[type="search"] {
 }
 
 .sage-btn--small {
-  @extend %t-sage-body-small-med;
-
   padding-top: rem(9px); // add 1px to offset odd line-height #
   padding-bottom: sage-spacing(xs);
+  font-size: sage-font-size(sm);
 }
 
 .sage-btn--full-width {
--- a/app/assets/stylesheets/sage/system/patterns/elements/_button.scss
+++ b/app/assets/stylesheets/sage/system/patterns/elements/_button.scss
@@ -68,12 +68,6 @@ input[type="search"] {
     opacity: 0;
   }
 
-  &:hover,
-  &:focus,
-  &:active {
-    color: sage-color(charcoal, 200);
-  }
-
   &:focus {
     box-shadow: $sage-btn-shadow-base;
     outline: none;
@@ -104,7 +98,7 @@ input[type="search"] {
 .sage-btn--small {
   @extend %t-sage-body-small-med;
 
-  padding-top: sage-spacing(xs);
+  padding-top: rem(9px); // add 1px to offset odd line-height #
   padding-bottom: sage-spacing(xs);
 }
 
--- a/app/assets/stylesheets/sage/system/patterns/elements/_button.scss
+++ b/app/assets/stylesheets/sage/system/patterns/elements/_button.scss
@@ -249,6 +249,16 @@ input[type="search"] {
   }
 }
 
+.sage-btn--link {
+  box-shadow: none;
+
+  &:focus,
+  &:active,
+  &:hover {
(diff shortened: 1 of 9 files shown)


philschanely (approved):

app/assets/stylesheets/sage/system/patterns/elements/_button.scss
```diff
@@ -49,9 +48,8 @@ input[type="search"] {
 .sage-btn {
   @include button-style-base;
 
-  // add space between adjacent buttons
-  + .sage-btn {
-    margin-left: sage-spacing(xs);
+  &:first-of-type {
+    margin-left: 0;
   }
```
AndrwM, line 53: Clever! 💯

app/assets/stylesheets/sage/system/patterns/elements/_button.scss
```diff
@@ -103,6 +95,16 @@ input[type="search"] {
...
 }
 
+.sage-btn--small {
+  padding-top: rem(9px); // add 1px to offset odd line-height #
+  padding-bottom: sage-spacing(xs);
+  font-size: sage-font-size(sm);
```
AndrwM: Could we use `sage-spacing(xs) + 1` instead?

Those line-height situations are so annoying some times. 🙃

app/views/sage/examples/objects/banner/_markup.html.erb
```diff
@@ -8,10 +8,10 @@
     </p>
   <% end %>
   <% if link and link_url %>
-    <a href="<%= link_url %>" class="<% if link_is_button %>sage-btn sage-btn--secondary sage-banner__btn <% end %>sage-banner__link">
+    <a href="<%= link_url %>" class="<% if link_is_button %>sage-btn sage-btn--small sage-btn--secondary sage-banner__btn <% end %>sage-banner__link">
       <%= link_text != "" ? " #{link_text}" : "link" -%>
     </a>
   <% elsif dismissable %>
-    <button class="sage-btn sage-banner__btn sage-banner__close"><%= dismiss_text != "" ? " #{dismiss_text}" : "" %></button>
+    <button class="sage-btn sage-btn--small sage-banner__btn sage-banner__close"><%= dismiss_text != "" ? " #{dismiss_text}" : "" %></button>
```
AndrwM: This doesn't follow the BEM convention where elements must be nested and can't be assigned roles in 2 different components.

AndrwM (changes requested):
The way the `sage-btn--small` is being used in the banner doesn't follow BEM conventions. I get this button implementation is a bit of an exception since it's on a solid color and therefore the shadow shows up really strong.

I think we cold probably solve this for now by removing the box-shadow for small buttons...though, not sure if this is a longterm solution.

Curious what the rest of the team thinks.

app/views/sage/examples/objects/banner/_markup.html.erb
teenwolfblitzer: Hmm, that's a good point. This one is a little tricky since the banner close button needs to be a `button` element, but removing the `box-shadow` on `btn--small` alone feels like an overreach. With that in mind, I've created a new class (`sage-btn--link`) intended for buttons styled to appear as inline text links. I believe is what the `link-button` element was originally intended for (but never completed) and can be removed.

AndrwM (approved):
Thanks for addressing this @monkeypox8 🙌

PR #191, opened 2020-04-22
adding spacer utility

cmcfadzean:
Adds a helper to be able to group items and add a martin at the bottom.


app/views/sage/examples/utilities/spacer/_preview.html.erb
```diff
@@ -0,0 +1,2 @@
+<!-- Can be wrapped around any group of elements to add consistant margin-bottom -->
+<div class="sage-spacer"></div>
```
philschanely, line 2: Any way we could do this without requiring a wrapper `div` such as doing this through an `::after` or `::before` on the element such as:

`<div class="sage-media-or-whatever sage-spacer--sm-after">...</div>`

Then of course:

```scss
.sage-spacer--sm-after {
  &::after {
    // sm gap settings
  }
}

```

I'm even ok with using an `hr` with `aria-hidden="true"` and no border but just height or margin.

Just hoping to avoid superfluous or redundant nesting wherever we can.

philschanely (commented):
Some thoughts on the approach.

app/views/sage/examples/utilities/spacer/_preview.html.erb
cmcfadzean: This is giving me a great thought. Going to do some rethinking and update this PR.

cmcfadzean:
Closing this as we really dont need this for now.

PR #195, opened 2020-04-24
Jcm sage 12.1

cmcfadzean:
An issue with showing the version number so removing it and bumping the version number as a bug fix. I will be testing on sage_public before we update on kajabi-products


pixelflips (approved):
👍 +1

PR #197, opened 2020-04-27
Convert switch markup to .erb

pixelflips:
- [x] Create markup file for the switch.


app/views/sage/examples/elements/switch/_markup.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="sage-switch">
+        <input id="<%= id %>" type="<%= type %>" class="sage-switch__input<%= has_error ? " sage-switch--error" : "" %>" name="<%= name %>" value="<%= id %>"<%= checked  ? " checked" : "" %><%= disabled ? " disabled" : "" %><%= required ? " required" : "" %>>
+        <label for="<%= id %>" class="sage-switch__label"><%= label_text %></label>
+        <% if message != "" %>
+        <div class="sage-switch__message">Additional Info</div>
+        <% else %>
+        <% end %>
+    </div>
```
AndrwM, line 8: Indention should be 2 spaces

AndrwM: Indention should be 2 spaces

app/views/sage/examples/elements/switch/_preview.html.erb
```diff
@@ -1,73 +1,172 @@
...
+    type: "checkbox",
+    name: "sage-switch-1",
+    id: "sage-switch-1",
+    value: "switch-1-value",
+    label_text: "Switch (checkbox)",
+    has_error: false,
+    checked: false,
+    disabled: false,
+    required: false,
+    message: ""
```
AndrwM: These attributes should be indented in

app/views/sage/examples/elements/switch/_markup.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="sage-switch">
+        <input id="<%= id %>" type="<%= type %>" class="sage-switch__input<%= has_error ? " sage-switch--error" : "" %>" name="<%= name %>" value="<%= id %>"<%= checked  ? " checked" : "" %><%= disabled ? " disabled" : "" %><%= required ? " required" : "" %>>
+        <label for="<%= id %>" class="sage-switch__label"><%= label_text %></label>
+        <% if message != "" %>
+        <div class="sage-switch__message">Additional Info</div>
+        <% else %>
+        <% end %>
```
AndrwM: `<% else %>` isn't needed here

app/views/sage/examples/elements/switch/_markup.html.erb
```diff
@@ -0,0 +1,8 @@
+<div class="sage-switch">
+        <input id="<%= id %>" type="<%= type %>" class="sage-switch__input<%= has_error ? " sage-switch--error" : "" %>" name="<%= name %>" value="<%= id %>"<%= checked  ? " checked" : "" %><%= disabled ? " disabled" : "" %><%= required ? " required" : "" %>>
```
AndrwM: Could we make these html attributes multi-line? The line length here is really long.

AndrwM (commented):
Awesome thanks for doing this! Left some nits regarding indention, aside from that it's a 👍 from me. 🥳

app/views/sage/examples/elements/switch/_markup.html.erb
pixelflips: @andrew - this might be a lack of ruby knowledge on my side, but I noticed without the empty else, it leave a blank line in the preview which makes copy/past have an empty line as well. Do you know if there is another way to possibly prevent that?

app/views/sage/examples/elements/switch/_markup.html.erb
pixelflips: > Could we make these html attributes multi-line? The line length here is really long.

@AndrwM I feel that it makes the preview much cleaner (visually) if these appear on 1 line. I agree it might be a bit long, but since we will only be editing the preview files it might be worth leaving them on one line for ease of readability and ease of copy/pasting. Any thoughts?

philschanely (approved):

PR #200, opened 2020-04-28
Sage-Select

### LINKED ISSUE #198: Sage Form Select

AndrwM:
Resolves: #198

# _Sage Select_ 🔮

- Adds sage-select element
- Fixes `position()` mixin (I broke it in my last PR)

### Notes
`<select>` doesn't have an empty/placeholder pseudo selector so I'm using `:invalid` to adjust the positioning of the label.

Feedback on the variable structure requested. The structure in this PR is more in-line with a "private variable" pattern. These variable will become fully "private" once we start using `@use`.

### Preview

screencast 2020-04-28 09-55-57

### DIFF
--- a/app/assets/stylesheets/sage/system/core/_mixins.scss
+++ b/app/assets/stylesheets/sage/system/core/_mixins.scss
@@ -132,7 +132,7 @@
   } @else if $value != unset {
     @return $value;
   } @else {
-    @return false;
+    @return null;
   }
 }
 
@@ -150,7 +150,7 @@
 ================================================== */
 
 @mixin icon-base($icon: null) {
-  display: inline-block;
+  display: inline-flex;
   text-transform: none;
   font: normal normal normal 1rem/1 "Sage";
   speak: none;
--- a/app/assets/stylesheets/sage/system/patterns/elements/_form_select.scss
+++ b/app/assets/stylesheets/sage/system/patterns/elements/_form_select.scss
@@ -32,8 +32,8 @@ $-sage-selectfield-color-error:               $sage-inputfield-color-error;
 .sage-select__label {
   position: absolute;
   white-space: nowrap;
-  pointer-events: none;
   line-height: initial;
+  pointer-events: none;
 
   @include position((($-sage-selectfield-height / 2) - $sage-body-font-size), unset, unset, $-sage-selectfield-padding-x);
 }
@@ -41,12 +41,11 @@ $-sage-selectfield-color-error:               $sage-inputfield-color-error;
 %sage-select__label--active {
   @extend %t-sage-body-xsmall-semi;
 
-  padding: 0 $-sage-selectfield-padding-label;
-  color: inherit;
-  background-color: $-sage-selectfield-bg-label;
   transform: translateY(calc(-50% - (#{$sage-body-font-size} / 4)));
-  transition: transform 0.15s ease-in-out, color 0.15s ease-out;
+  padding: 0 $-sage-selectfield-padding-label;
   color: $-sage-selectfield-color-success;
+  background-color: $-sage-selectfield-bg-label;
+  transition: transform 0.15s ease-in-out, color 0.15s ease-out;
 }
 
 %sage-select__label--inactive {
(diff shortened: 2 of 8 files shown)


app/assets/stylesheets/sage/system/core/_mixins.scss
```diff
@@ -132,7 +132,7 @@
   } @else if $value != unset {
     @return $value;
   } @else {
-    @return false;
+    @return null;
```
philschanely, line 135: 😍 Good catch here.

app/assets/stylesheets/sage/system/patterns/elements/_form_select.scss
```diff
@@ -0,0 +1,130 @@
...
+.sage-select__label {
+  position: absolute;
+  white-space: nowrap;
+  pointer-events: none;
+  line-height: initial;
+
+  @include position((($-sage-selectfield-height / 2) - $sage-body-font-size), unset, unset, $-sage-selectfield-padding-x);
+}
+
+%sage-select__label--active {
```
philschanely, line 41: Is this modularization necessary? Should we perhaps have a centralized label mixin? Am I missing something?

app/assets/stylesheets/sage/system/patterns/elements/_form_select.scss
```diff
@@ -0,0 +1,130 @@
...
+  color: transparent;
+  border: $-sage-selectfield-border-width solid $-sage-selectfield-border-color;
+  border-radius: $-sage-selectfield-border-radius;
+  background: transparent;
+  outline: none;
+  cursor: pointer;
+
+  // The <Select> is set as [required] & therefore is invalid when the value
+  // is empty, we're mapping :invalid to "empty" for more legible code.
+  $empty: "invalid";
```
philschanely, line 86: 🥇

philschanely (approved):
Nice work. Just a non-blocking question.

app/assets/stylesheets/sage/system/patterns/elements/_form_select.scss
```diff
@@ -0,0 +1,130 @@
...
+  line-height: initial;
+
+  @include position((($-sage-selectfield-height / 2) - $sage-body-font-size), unset, unset, $-sage-selectfield-padding-x);
+}
+
+%sage-select__label--active {
+  @extend %t-sage-body-xsmall-semi;
+
+  padding: 0 $-sage-selectfield-padding-label;
+  color: inherit;
```
teenwolfblitzer: Was the duped value here intentional? Looks like stylelint is choking on this one.

app/assets/stylesheets/sage/system/patterns/elements/_form_select.scss
AndrwM: Yeah, I think we should. Good idea. Right now we're using this style within:
- textarea
- dropdown
- string input
- select

There's some variation in how the positioning needs to be handled in some of the components but across the board it's the same visual style.

AndrwM: The verdict here: Shipping this as-is and refactoring in a followup PR.

app/assets/stylesheets/sage/system/patterns/elements/_form_select.scss
```diff
@@ -0,0 +1,130 @@
...
+  border-radius: $-sage-selectfield-border-radius;
+  background: transparent;
+  outline: none;
+  cursor: pointer;
+
+  // The <Select> is set as [required] & therefore is invalid when the value
+  // is empty, we're mapping :invalid to "empty" for more legible code.
+  $empty: "invalid";
+
+  &:hover:#{$empty} {
```
teenwolfblitzer, line 89: Could we also add on `:focus` to this (and the style below?)

teenwolfblitzer (commented):
Couple of small items to check out, looking great so far!

I did run into a weird bug with Firefox where the dropdown text becomes invisible. It didn't happen consistently, so it could just be my system.:
|  default | error  |
|--------|--------|
|ff-select|ff-select2|

app/assets/stylesheets/sage/system/patterns/elements/_form_select.scss
AndrwM: Oops, nope! Thanks for catching

AndrwM:
@monkeypox8 Was able to isolate that bug. TIL, firefox allows you to set the color of `<option>`. It was inheriting `transparent` from `<select>`.

Addressed this and the other issues above. Thanks!

teenwolfblitzer (approved):
👍 🔥 Looks great!

PR #204, opened 2020-04-28
Convert form input markup to .erb

pixelflips:
- [x] Create markup file for the form input.


philschanely (approved):

app/views/sage/examples/elements/form_input/_markup.html.erb
```diff
@@ -0,0 +1,5 @@
+<div class="sage-input<%= has_error ? " sage-input--error" : "" %>">
+  <input type="<%= input_type %>" id="<%= id %>" class="sage-input__field" name="<%= id %>" placeholder="<%= placeholder %>" <%= required ? "required" : "" %><% if input_mode != "" %> inputmode="<%= input_mode %>"<% end %><% if max != "" %> max="<%= max %>"<% end %><% if minlength != "" %> minlength="<%= minlength %>"<% end %><% if maxlength != "" %> maxlength="<%= maxlength %>"<% end %>>
```
AndrwM, line 2: You can also write these `if` statements like this:
```
<%= "minlength='#{minlength}'" if minlength.present? %>
```

This approach seems a bit less verbose to me, up to you on which you prefer.

AndrwM (approved):
Looks good! Thanks +1

app/views/sage/examples/elements/form_input/_markup.html.erb
pixelflips: @AndrwM - thanks so much I like it! I will adjust.

PR #207, opened 2020-04-28
Convert danger button markup to .erb

### LINKED ISSUE #209: Consolidate "danger" button styles into parent `sage-btn`
- [ ] Remove danger button markup and element references
- [ ] Consolidate CSS
- [ ] Update `sage-btn` example and markup

pixelflips:
- [x] Create markup file for the danger button.

### DIFF
--- a/app/views/sage/examples/elements/danger_button/_markup.html.erb
+++ b/app/views/sage/examples/elements/danger_button/_markup.html.erb
@@ -1,0 +1,1 @@
+<<%= element %><% if element != "button" %> href="#"<% end %> class="sage-btn sage-btn--danger<% if element != "button" && disabled %> disabled<%end%>"<% if element != "a" && disabled %> disabled<%end%><% if aria_disabled %> aria-disabled="true"<% end %>><%= text %></<%= element %>>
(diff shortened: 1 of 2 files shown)


pixelflips:
@cmcfadzean - since this one is slightly different than the other conversions (there are two different possible elements) I wanted to double-check that the way I am generating the different elements is ok? Basically, I am passing in the element from the preview in order for it to create the link or button.

Any thoughts or possible alternative approaches would be greatly appreciated before I invite everyone for a general review. Thanks!

cc: @monkeypox8

teenwolfblitzer:
While this looks great, the danger button was intended to be folded back into the `sage-btn` parent at some point, since there isn't a need for them to live as separate elements. `sage-btn--danger` should really be all we need. That being said, the same template generation logic you've created here may also work just as well for the parent markup once we're ready to take that on.

pixelflips:
@monkeypox8 - ok that makes sense. With that said, I think I will go ahead with the PR (since it's done) and then we can just kill off the danger button all together once we get to the button conversion/updates. Think that works?

teenwolfblitzer (approved):
👍LGTM, nice work!

Issue #184, opened 2020-04-20
Table styles


teenwolfblitzer:
Revise table styles for use in admin & superadmin
- [x] Table header sort indicators - current sort visibility, sortable column link appearance (#244)
- [x] Column spacing and alignment (#186)
- [x] Show/hide columns (#252)
- [ ] ~Embedded button styles - TBD~
- [ ] Filtering and search — FUTURE (final implementation TBD)

teenwolfblitzer:
Primary tasks are complete. Closing this one out in favor of #255 for extended work.

PR #186, opened 2020-04-29
Table column options

teenwolfblitzer:
## Description
Update features of Tables using work in Superadmin (see https://github.com/Kajabi/kajabi-products/pull/14131)

Follow-up to come addressing showing/hiding individual columns a la Bootstrap.

## Related issues
- part of #184

### DIFF
--- a/app/assets/stylesheets/sage/system/patterns/elements/_table.scss
+++ b/app/assets/stylesheets/sage/system/patterns/elements/_table.scss
@@ -77,6 +77,18 @@ $-sage-table-cell-truncate-width: 6.5em;
   }
 }
 
+.sage-table--condensed {
+  thead,
+  tfoot,
+  tbody {
+    td,
+    th {
+      padding-top: $sage-table-cell-padding-xs;
+      padding-bottom: $sage-table-cell-padding-xs;
+    }
+  }
+}
+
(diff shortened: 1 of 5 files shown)


pixelflips (approved):
👍 +1 LGTM!

app/assets/stylesheets/sage/system/patterns/elements/_table.scss
```diff
@@ -71,14 +77,48 @@
...
     }
   }
 }
 
+
+// Selectable rows with checkbox or radio inputs
+.sage-table--selectable {
+  .sage-row--selectable,
+  .sage-row--selectable:nth-child(odd) {
+    transition: background 0.1s ease-in-out;
```
philschanely: Is this anywhere a consistent setting that we'd want in a variable? I've separated the transition-property out in some cases.

philschanely (approved):

app/assets/stylesheets/sage/system/patterns/elements/_table.scss
teenwolfblitzer: Oh, great point. I should be using the global `$sage-transition` here. I'll make that change.

PR #208, opened 2020-04-29
Webpacker Prototype Branch

AndrwM:
#### NOTE: DEPENDS ON [KAJABI PRODUCTS PR #14378](https://github.com/Kajabi/kajabi-products/pull/14378)
----

# Webpacker Migration Release Branch 🏜
Splitting this into 2 followup PRs for clean diffs. The PRs below will be merged into this one.
This PR is the initial work of webpacker conversion and integration into kajabi-products.

👉 **Once this PR has been vetted and approved the followup PRs will go into review.**

---

### Webpacker Setup & Products Consumption → Current Branch
- [x] Migrate from NPM to Yarn (dependancy of webpacker)
- [x] Successfully compile packs
- [x] test/dummy pack consumption
- [x] ES6 transpilation with Babel
- [x] Auto reloading with webpacker (run dev-server in tandem with rails server)
- [x] ~Production compilation rake tasks~
- [x] ~Setup middleware injection for kajabi-products to accept the js~
- [x] Update package.json scripts with new implementation
- [x] Use `./tmp/sage/packs` instead of `./public/sage/packs`
- [x] ~Webpack clobber rake task~
- [x] Migrate stylelint implementation (commented for now, was having failures in the mounted engine)
- [x] [Development workflow (documentation) for kajabi-products with a local Sage gem →](https://github.com/Kajabi/sage/blob/7ae420ffa80ec27985290762f7ac416bc9a7638d/bin/sage-local-link.sh)
- [ ] Cache-bust node_module file changes in Products when using local Sage

---

### 2/3: Migrate Assets To Webpack → #223 
- [x] Migrate JS to webpack
- [x] Migrate CSS to webpack
- [ ] ~Code Prettify~ (this package is deprecated and isn't built to be ES `import`ed)

---

### 2/3: Cleanup & Tooling
- [x] Remove the old assets and references
- [ ] Remove unnecessary asset compilation gems (autoprefixer, etc)
- [ ] ~Auto Sage webpack precompile when not using the local sage gem (if necessary)~
- [ ] Update generators

-----

### Sage Dummy App
```
yarn start
```

### Kajabi-Products
Run local setup within `kajabi-products`
```
<PATH TO SAGE REPO>/bin/sage-local-link true
```
When running locally within Kajabi-Products you may need to run `rake webpacker:clobber` occasionally so that Webpacker reloads the local gem. (A fix is incoming for this.)


AndrwM:
@jacobthemyth This is ready for your eyes again!

jacobthemyth:
👀

app/helpers/sage/webpacker_helper.rb
```diff
@@ -0,0 +1,10 @@
+module Sage
+  module WebpackerHelper
+    include ::Webpacker::Helper
+
+    def current_webpacker_instance
```
jacobthemyth, line 5: Will this still be used when consuming as a node module? This might be part of your cleanup, I'm just curious if there will still be cross talk between webpacker instances.

jacobthemyth (commented):
Is the idea that sage would be in both `Gemfile` and `package.json`? I figured it would only be in `package.json` so that much of the Ruby stuff could go away, but I don't think I'm quite seeing how everything fits together. Having it in the `Gemfile` means that we'll have to `Webpacker` instances I think, though I think you've got that working. If it was only in `package.json`, the compiled files would need to be added to the repo instead of in `tmp/sage` I think though? Anyway, this seems like a legit approach, I'm not sure sure I fully understand why it's still a Ruby gem. Happy to get on a call if that would help.

app/helpers/sage/webpacker_helper.rb
AndrwM: Thanks for looking through this!

The idea for using the Sage gem in tandom with a package implementation is that partialized UI patterns can be migrated into the Sage gem to silo the UI from Kajabi-Products. We're doing this already within Kajabi-Products at the moment (see `app/sage_components/` & `app/views/sage_components`).

At the moment outside of providing assets to AssetPipeline (which we are now no longer doing as of this PR) we're just mounting this engine to the route `/sage` in development. The engine provides all the documentation for the system.

The above code ensures we're using our Sage gem engine's webpacker instance within both our `test/dummy` app (our local documentation development enviroment) and within the Kajabi-Products development-mounted engine.

Though, at the moment I'm not sure how many devs use the documentation mounted within Kajabi-Products over our hosted version here (https://sage-design-system.kajabi.com).

I like the idea of simplifying this more, it seems a bit unconventional that we would include both a package and a gem. I'm curious what kinds of solutions would be good in this instance.

I'll reach out over Slack to schedule a time to chat 😄

AndrwM: This implementation follows the suggested Engine setup here: https://github.com/rails/webpacker/blob/master/docs/engines.md

AndrwM:
Converting this to a draft so that it doesn't show on my task list.

AndrwM:
Closing this branch in favor of merging:
https://github.com/Kajabi/sage/pull/274

☝️  child branch of this branch.

PR #210, opened 2020-04-29
Convert live stream wrapper markup to .erb

pixelflips:
- [x] Create a markup file for the live stream wrapper.


philschanely (approved):
Thanks for this!

PR #211, opened 2020-04-29
Convert table markup to .erb

pixelflips:
- [x] Create table markup file
- [x] Update markup and preview files


philschanely (approved):
Would you consider allowing this to accept a collection that it parses to rows and cells? Not blocking this implementation though.

pixelflips:
> Would you consider allowing this to accept a collection that it parses to rows and cells? Not blocking this implementation though.

@philschanely - sure! care to elaborate though so I am sure I understand. I think you mean possibly being able to pass in data for the column headers as an example?

philschanely:
Yeah something like that; some form of being able to call the partial and pass in a data set, rather than the partial always showing the same fixed data. No biggie.

philschanely:
And by "no biggie" I mean its ok if its something we put in a separate task.

PR #212, opened 2020-04-30
JS error on products/superadmin

teenwolfblitzer:
While working in the superadmin, I noticed that the Sage JS did not appear to be initializing, though the global `Sage` object was accessible from the console. The solution being to add the inline `init` script (`application/_scripts.html.erb`), which seems to be missing from the page templates. During troubleshooting, @cmcfadzean pointed out that there was an error on prod, which this PR should correct.

- Corrects issue with JS variable initializing incorrectly
- Revises loading strategy for banner JS


pixelflips (approved):
👍+1 - Nice work!

PR #213, opened 2020-04-30
Version bump to `0.13.0`

teenwolfblitzer:
⬆️ 

### Added
* New Billboard Object
* Rails template markup examples added for elements:
  - Radio
  - Switch
  - Label
  - Form input
  - Form textarea
  - Loader
  - Danger button

### Updated/Fixed
* Updated the panel object to having both a header and a footer 
* Fixed error and refactored `banner.js`


pixelflips (approved):
👍 +1

PR #215, opened 2020-05-01
Icon sizing in buttons

### LINKED ISSUE #216: Consolidate danger button into parent `sage-btn`
Danger button should no longer be a separate element and can be folded into the main `sage-btn`. CSS class has already been consolidated.

teenwolfblitzer:
## Description
Icons in buttons when using `sage-btn--small` are out of alignment with text due to a smaller font-size. One possible solution would be to adjust the font-size in the icon generator to use `1em` instead of `1rem`.

This PR also addresses consolidating the separate danger button markup/element into the main button.

|  `sage-btn--primary`   |  `sage-btn--secondary`  |
|--------|--------|
|btn-primary|btn-secondary|

## Related
- Resolves #216


app/assets/stylesheets/sage/system/core/_mixins.scss
```diff
@@ -152,7 +152,7 @@
 @mixin icon-base($icon: null) {
   display: inline-flex;
   text-transform: none;
-  font: normal normal normal 1rem/1 "Sage";
+  font: normal normal normal 1em/1 "Sage";
```
teenwolfblitzer, line 155: This change may mean we'll have to reconsider scenarios with unwanted icon scaling. Previously we prevented this behavior by using a fixed size (with a future goal of creating a defined scale).

pixelflips (approved):
👍 +1 - LGTM!

philschanely (approved):
No concerns here per se but I do believe Design is going to prescribe specific icon sizes at some point.

PR #219, opened 2020-05-01
Convert card markup to .erb

pixelflips:
- [x]  Create a markup file for card.

### DIFF
--- a/app/views/sage/examples/objects/card/_markup.html.erb
+++ b/app/views/sage/examples/objects/card/_markup.html.erb
@@ -1,0 +1,10 @@
+<div class="sage-card<%= is_compact ? " sage-card--compact" : "" %><%= is_spaced ? " sage-card--spaced" : "" %>">
+      <% if has_img %>
+      <img class="sage-card__img" src="<%= img_path %>" alt="">
+      <% end %>
+      <div class="sage-card__body">
+        <h2 class="sage-card__title t-sage-heading-5"><%= title %></h2>
+        <p class="sage-card__text"><%= text %></p>
+        <button class="sage-btn <% if btn_style_primary %> sage-btn--primary <% elsif btn_style_tertiary %> sage-btn--tertiary sage-btn--icon-right-caret-right<% end %>"><%= btn_text %></button>
+      </div>
+    </div>
--- a/app/views/sage/examples/objects/card/_markup.html.erb
+++ b/app/views/sage/examples/objects/card/_markup.html.erb
@@ -1,10 +1,10 @@
 <div class="sage-card<%= is_compact ? " sage-card--compact" : "" %><%= is_spaced ? " sage-card--spaced" : "" %>">
-      <% if has_img %>
-      <img class="sage-card__img" src="<%= img_path %>" alt="">
-      <% end %>
-      <div class="sage-card__body">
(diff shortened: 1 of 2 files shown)


app/views/sage/examples/objects/card/_markup.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="sage-card<%= is_compact ? " sage-card--compact" : "" %><%= is_spaced ? " sage-card--spaced" : "" %>">
+      <% if has_img %>
+      <img class="sage-card__img" src="<%= img_path %>" alt="">
```
teenwolfblitzer: We probably would want to add an option to enter `alt` text for this image.

app/views/sage/examples/objects/card/_markup.html.erb
```diff
@@ -0,0 +1,10 @@
...
+      <% end %>
+      <div class="sage-card__body">
+        <h2 class="sage-card__title t-sage-heading-5"><%= title %></h2>
+        <p class="sage-card__text"><%= text %></p>
+        <button class="sage-btn <% if btn_style_primary %> sage-btn--primary <% elsif btn_style_tertiary %> sage-btn--tertiary sage-btn--icon-right-caret-right<% end %>"><%= btn_text %></button>
```
teenwolfblitzer: I wonder if this element should be optional and/or an anchor tag rather than a button, since presumably it'll be primarily used as a link to other pages.

I'd say make it an anchor if we can for now, the rest can be a follow-up.

app/views/sage/examples/objects/card/_markup.html.erb
```diff
@@ -0,0 +1,10 @@
+<div class="sage-card<%= is_compact ? " sage-card--compact" : "" %><%= is_spaced ? " sage-card--spaced" : "" %>">
+      <% if has_img %>
```
teenwolfblitzer: Super nitpicky comment: the indentation (spacing) seems off here, as if lines 2+ are indented too far in.

teenwolfblitzer (commented):
Looking great, a few items to consider modifying.

app/views/sage/examples/objects/card/_markup.html.erb
pixelflips: @monkeypox8 - makes sense, I'll make it a link.

app/views/sage/examples/objects/card/_markup.html.erb
pixelflips: @monkeypox8 you're right and not nitpicky at all. I did notice that when I fix the indentation as expected, it shows up a bit strange in the preview. Is the preview something we can adjust? I'll let you know when I have pushed these adjustments and you can have a look.

philschanely (approved):

app/views/sage/examples/objects/card/_markup.html.erb
teenwolfblitzer: Good eye. It seems to happen in all of the previews we've updated, but I don't think it's something that we can alter without affecting readability in both the preview and markup due to the fact that `pre` will render all spaces. It seems to be a combination of a) Ruby outputting the partial/include with whitespace intact, and b) our own indentation for readability, even when using whitespace [control tags](https://api.rubyonrails.org/classes/ActionView/Base.html).

PR #179, opened 2020-04-23
Revise alert styling and documentation

pixelflips:
## Description
Update alert styling, functionality, and documentation to account for icons and dismissable actions. 

- Previously named as "banners", which have been superseded by the new banner object
- **NOTE:** This update does not include updated icons and uses the existing icons.

THIS UPDATE REQUIRES AN UPDATE IN KAJABI-PRODUCTS DUE TO RENAMING.

alerts

## Related issues
- #176

PR #230, opened 2020-05-05
Update alert so closing is optional

pixelflips:
## Description
In recent updates to the alert object (#179), a close X was added but always appears.

This update will include minor adjustments so that the close X can be optional to support the use case for when an alert should remain persistent.


philschanely (approved):
👍

teenwolfblitzer (approved):
👍 nice work!

PR #232, opened 2020-05-05
Convert nav markup to .erb

pixelflips:
- [x] Create a markup file for nav.

### DIFF
--- a/app/views/sage/examples/objects/nav/_markup.html.erb
+++ b/app/views/sage/examples/objects/nav/_markup.html.erb
@@ -1,5 +1,5 @@
-<div class="sage-nav">
+<nav class="sage-nav" aria-label="Main">
   <% nav_links.each do |link| %>
-  <a href="#" class="sage-nav__link <%= link[:is_sublink] ? "sage-nav__link--sub" : "" %>"><%= link[:text] %></a>
+  <a href="<%= link[:url] %>" class="sage-nav__link <%= link[:is_sublink] ? "sage-nav__link--sub" : "" %>"><%= link[:text] %></a>
   <% end %>
-</div>
+</nav>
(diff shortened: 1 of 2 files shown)


app/views/sage/examples/objects/nav/_markup.html.erb
```diff
@@ -0,0 +1,5 @@
+<div class="sage-nav">
```
teenwolfblitzer: Hmm, not sure why this wasn't the case before, but this `div` should be a `nav` element with an `aria-label` indicating the region (see the actual sidebar for an example). Those two items will help distinguish the landmark in screen readers.

Ideally, we'd also use an unordered list grouping the nav items together, but that's better addressed in a follow-up.

app/views/sage/examples/objects/nav/_markup.html.erb
```diff
@@ -0,0 +1,5 @@
...
+  <% nav_links.each do |link| %>
+  <a href="#" class="sage-nav__link <%= link[:is_sublink] ? "sage-nav__link--sub" : "" %>"><%= link[:text] %></a>
```
teenwolfblitzer: Should we also make the link URLs here editable in the options?

teenwolfblitzer (commented):
One item to address and one to consider, nice work so far!

app/views/sage/examples/objects/nav/_markup.html.erb
pixelflips: @monkeypox8 thanks for the info. I'll update the element to nav and add the `aria-label` and then I'll create an issue in the backlog for the `ol` update.

app/views/sage/examples/objects/nav/_markup.html.erb
pixelflips: Sure. Will adjust.

teenwolfblitzer (approved):
👍 🔥 Nice work!

PR #233, opened 2020-05-05
HTML Validation cleanup

teenwolfblitzer:
## Description
NOTE: # of lines changed is actually misleading, the majority is from a whitespace/indentation change in `icon.html.erb` 🤷‍♂️ 

Cleans up some errors noted while running validation and accessibility testing. Also removes a few duplicate examples of the switch element.

One item we may be unable to fix is the use of `cache="true"` on scripts generated by the Rails asset pipeline. Keeps us from being 100% valid HTML, but not a breaking issue.

Several additional errors/warnings in the following components regarding roles and `aria` labels will require a follow-up:
- `sage-live-user-control`
- `sage-live-profile-card`
- `sage-live-stream-control`

## Related
- #231 

Full report available here:
[jsonformatter-2020-05-05-13-48-09-.json.txt](https://github.com/Kajabi/sage/files/4583556/jsonformatter-2020-05-05-13-48-09-.json.txt)


pixelflips (approved):
👍 - Nice work and thanks for cleaning up!

philschanely (approved):

PR #234, opened 2020-05-06
Convert sidebar markup to .erb

pixelflips:
- [x] Create a markup file for the sidebar.
- [x] Update nav and render it in the sidebar
- [x] Add active option to nav object (for active styling)


teenwolfblitzer (approved):
👍 Looks great! For a follow-up, I'm thinking we take it one step further:

```
<div class="sage-sidebar" data-js-sidebar="<%= sidebar_id %>" id="<%= sidebar_id %>">
  <div class="sage-sidebar__body">
    <%- if body_content -%>
      <%= render body_content -%>
    <%- end -%>
  </div>
  <div class="sage-sidebar__footer">
    <%- if footer_content != "" -%>
      <%= render footer_content -%>
    <%- end -%>
  </div>
</div>
```
Then we can just pass the string `"sage/examples/objects/nav/preview"` to render the nav rather than hard-coding it into the template.

pixelflips:
@monkeypox8 that could work, but do you imagine a scenario where there would be no `body ` content? Meaning do we need the condition for the body? We could still pass in the `body_content`, but just curious as to why check if it exists?

I could see that for the footer possibly, but the sidebar without a body is fairly useless, no?

teenwolfblitzer:
> @monkeypox8 that could work, but do you imagine a scenario where there would be no `body ` content? Meaning do we need the condition for the body? We could still pass in the `body_content`, but just curious as to why check if it exists?

That's an excellent point, we probably don't need the extra conditional considering it'll be developers consuming this and not consumers. No need to foolproof it quite as much.

pixelflips:
> That's an excellent point, we probably don't need the extra conditional considering it'll be developers consuming this and not consumers. No need to foolproof it quite as much.

@monkeypox8 Ok with that said, let's not wait for a follow-up. I'll adjust it in this PR and just ask for another review when it's ready. Thanks for all the help and suggestions!

teenwolfblitzer (approved):
🚢 🔥Ship it!

PR #239, opened 2020-05-06
Convert page heading markup to .erb

pixelflips:
- [x]  Create a markup file for page heading object.

### DIFF
--- a/app/views/sage/examples/objects/page_heading/_markup.html.erb
+++ b/app/views/sage/examples/objects/page_heading/_markup.html.erb
@@ -1,0 +1,7 @@
+<div class="sage-page-heading">
+  <a href="#" class="sage-page-heading__back">
+    <i class="sage-page-heading__icon sage-icon-caret-left" aria-hidden="true"></i>
+    <%= back_text %>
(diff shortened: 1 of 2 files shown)


philschanely (approved):

app/views/sage/examples/objects/page_heading/_markup.html.erb
```diff
@@ -0,0 +1,7 @@
+<div class="sage-page-heading">
+  <a href="<%= back_url %>" class="sage-page-heading__back">
```
teenwolfblitzer, line 2: Would there ever be a case where the back link might be omitted here?

teenwolfblitzer (approved):
👍 Looks good to me! One question, not a blocker to merge.

app/views/sage/examples/objects/page_heading/_markup.html.erb
pixelflips: @monkeypox8 Good question. I am not totally sure. I'll merge this update for now and make myself a note to find out if it should be optional.

PR #240, opened 2020-05-06
Convert panel markup to .erb

pixelflips:
- [x]  Create a markup file for the panel object.

### DIFF
--- a/app/views/sage/examples/objects/panel/_markup.html.erb
+++ b/app/views/sage/examples/objects/panel/_markup.html.erb
@@ -1,5 +1,5 @@
 <div class="sage-panel">
   <div class="sage-panel__header"><%= header %></div>
-  <div class="sage-panel__body"><%= body %></div>
+  <div class="sage-panel__body"><%= body.html_safe %></div>
   <div class="sage-panel__footer"><%= footer %></div>
 </div>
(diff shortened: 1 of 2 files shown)


app/views/sage/examples/objects/panel/_markup.html.erb
```diff
@@ -0,0 +1,5 @@
+<div class="sage-panel">
+  <div class="sage-panel__header"><%= header %></div>
+  <div class="sage-panel__body"><%= body %></div>
```
philschanely: Is there anything we need to to here for the change that the `body` has HTML markup in it?

philschanely (approved):
One side comment non-blocking.

app/views/sage/examples/objects/panel/_markup.html.erb
pixelflips: @philschanely - great question and didn't really think it through since I was just working on the conversion. I believe ruby would allow us to do something like `<%= body.html_safe %>` which would then allow us to pass a string containing HTML markup and it seems to display properly.

@cmcfadzean - any thoughts here and is this the type of approach the way to go?

pixelflips: @philschanely - nice catch and I've confirmed with Court that `<%= body.html_safe %>` will solve the issue so we would be able to pass in a plain 'ol string or HTML markup. I'll push an update.

PR #243, opened 2020-05-07
Convert pagination markup to .erb

pixelflips:
- [x] Create a markup file for the pagination object.

### DIFF
--- a/app/assets/stylesheets/sage/system/patterns/objects/_pagination.scss
+++ b/app/assets/stylesheets/sage/system/patterns/objects/_pagination.scss
@@ -7,6 +7,7 @@
 .sage-pagination {
   display: flex;
   justify-content: flex-end;
+  flex-flow: row wrap;
 }
 
 .sage-pagination__count {
@@ -17,6 +18,7 @@
 .sage-pagination__link {
   @extend %t-sage-body-semi;
   display: flex;
+  flex-flow: row wrap;
   align-items: center;
   margin: 0 sage-spacing(2xs);
   color: sage-color(primary);
@@ -35,6 +37,7 @@
 
 .sage-pagination__pages {
   display: flex;
+  flex-flow: row wrap;
   padding-left: 0;
   list-style-type: none;
 }
--- a/app/assets/stylesheets/sage/system/patterns/objects/_pagination.scss
+++ b/app/assets/stylesheets/sage/system/patterns/objects/_pagination.scss
@@ -18,11 +18,12 @@
   @extend %t-sage-body-semi;
   display: flex;
   align-items: center;
-  padding: 0 sage-spacing(xs);
+  margin: 0 sage-spacing(2xs);
   color: sage-color(primary);
 }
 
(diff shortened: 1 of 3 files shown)


app/views/sage/examples/objects/pagination/_markup.html.erb
```diff
@@ -0,0 +1,20 @@
+<nav class="sage-pagination" aria-label="Pagination">
+  <span class="sage-pagination__count"><strong><%= total_count %></strong> Records</span>
+  <a class="sage-pagination__link" rel="prev" href="<%= prev_link %>"><i class="sage-pagination__icon sage-icon-caret-left"></i> <%= prev_text %></a>
+  <ul class="sage-pagination__pages">
+    <% pagination_items.each do |item| %>
+    <li>
+      <%- if item[:is_link] -%>
+      <a href="<%= item[:link] %>" class="sage-pagination__page" aria-label="<%= item[:aria_label] %>"><%= item[:page_number] %></a>
```
teenwolfblitzer: One item to consider here, I think we could streamline the configuration by removing the `aria-label` and adding text in a `.visually-hidden` span. That way there's no need to duplicate the page number and text for each item in `pagination_items`, which is unlikely to change:

Something like `<span class="visually-hidden">Go to page </span><%= item[:page_number] %>`

Probably more of a reach here, but the same may also go for the "Prev" and "Next" links: I wonder if those are explanatory enough as standalone labels. Like the page links we could use the `.visually-hidden` class to hide some descriptive text.

teenwolfblitzer (approved):
👍 Looking great, couple of optional accessibility suggestions to consider.

app/views/sage/examples/objects/pagination/_markup.html.erb
pixelflips: @monkeypox8 - based on our conversation, I've made some adjustments. I'll request another review and feel free to have a look when you have a minute.

app/views/sage/examples/objects/pagination/_markup.html.erb
```diff
@@ -0,0 +1,23 @@
+<nav class="sage-pagination" aria-label="pagination">
+    <span class="sage-pagination__count"><strong><%= total_count %></strong> Records</span>
```
teenwolfblitzer: Small nitpick, but I think the indentation is off by one stop here

app/assets/stylesheets/sage/system/patterns/objects/_pagination.scss
```diff
@@ -18,11 +18,12 @@
   @extend %t-sage-body-semi;
   display: flex;
   align-items: center;
```
teenwolfblitzer, line 22: Could we add `flex-flow: row wrap` here? On small screens the flex container is too wide (screen attached) because of the "Records" content.

pagination

teenwolfblitzer (approved):
👍 🔥 Looks good to me! Couple of small non-blocking items to consider. Great work!

PR #244, opened 2020-05-08
Table column options

teenwolfblitzer:
## Description
Adding a few options here to match Superadmin table functionality and styles from Design.

- Sortable column styling and simple JS functionality (does not implement actual sort of tabular data)
- Updated documentation and examples

## Related
- #184


pixelflips (approved):
👍 +1 - LGTM! Nice job on updating the captions as well!

PR #246, opened 2020-05-08
Convert link_button markup to .erb

pixelflips:
- [x] Create a markup file for the form_select element.


pixelflips:
@monkeypox8 - I went ahead and converted this one as well, but will reach out to design later on to see if it's actually still needed. I think this is the last one!!

teenwolfblitzer (approved):
👍 I have a feeling we'll deprecate this one soon, but should be good to go for now.

PR #247, opened 2020-05-09
Shortened code previews

teenwolfblitzer:
## Description
I got a little tired of scrolling through the code previews to reach the properties while building out the `table` element, so I thought I'd try to clean them up. Code previews are shortened by default, with a button to expand when needed.

Applies only to the expanded element/object views and not on the list view.

### Screenshots
|  before  |  after  |
|--------|--------|
|before|after|

## Related
- n/a


pixelflips (approved):
👍 +1 LGTM - I really like this, nice work!

PR #248, opened 2020-05-11
Update card image path

pixelflips:
## Description
The card images are returning 404 on the Sage public site. The image path needs to be adjusted to correct this issue.

- [x] Convert image inclusion to `img_tag` instead of standard HTML `img` element.
- [x] Update file paths passed to each card.

### Screenshots
|  Before   |  After  |
|--------|--------|
|Screen Shot 2020-05-11 at 8 53 22 AM|Screen Shot 2020-05-11 at 8 53 32 AM| 

## Test notes
There should be no visual changes and this fix applies to the public site. Images were already appearing ok on local builds. The fix will need to be verified once the public site version is updated.


philschanely (approved):
ooo nice

PR #249, opened 2020-05-11
Correct image paths & optimizations

pixelflips:
## Description
While updating the image file paths for the cards, I found other areas where images were also not loading properly on the public sage site.

Converting the `img` elements to `img_tag` and correcting the file path being passed to the element or object fixes the issue.

- [x] live_avatar
- [x] live_profile_card
- [x] live_stream_video_grid

In addition, I've optimized and compressed the billboard and avatar images to reduce payload.

## Test notes
There should be no visual changes, except for images that should now load properly.


app/views/sage/examples/objects/live_avatar/_markup.html.erb
```diff
@@ -3,6 +3,6 @@
     <text text-anchor="middle" x="19.5" y="24"><%= initials %></text>
   </svg>
   <% if url != "" %>
-  <img src="<%= url %>" class="sage-live-avatar__image" alt="" />
+  <%= image_tag("#{url}", class: "sage-live-avatar__image", alt: "#{initials} Avatar") %>
```
teenwolfblitzer: This one's kind of a judgement call since the initials text isn't exactly descriptive, but as the image here is purely decorative, I'd recommend leaving the `alt` text empty/blank as it was before.

teenwolfblitzer (approved):
👍 Looking good, left a comment on the image `alt` text that I'll leave up to your judgement.

PR #104, opened 2020-05-12
Password masking

teenwolfblitzer:
## Description
- New component for upcoming work in Checkout, allows input field grouping with buttons (mainly icon buttons for now)
- Adds password input show/hide toggle functionality
- Fixes `:focus` state color issue on inputs with errors
- Password error state can be ignored at this time. Follow-up will implement pw strength tooltip and logic (prototype: https://codepen.io/monkeypox8/pen/RwWQQpo)

## Related
- #256

PR #261, opened 2020-05-15
Password tooltip

teenwolfblitzer:
## Description
A component that displays an expanded tooltip for text inputs to be used with upcoming work in the Checkout.

- used for a list of qualifications that must be met when creating a new password
- uses external library ([zxcvbn](https://github.com/dropbox/zxcvbn)) for password strength indicator
- can be grouped with Input Group for show/hide password functionality

input-helper

Relies on #263 merge

## Steps for testing
1. Navigate to "Input Group" in the list of Objects
2. Click or focus in the first example, "Input group with password reveal toggle and helper"
3. Start typing

## Related

- #104


app/assets/javascripts/sage/system/inputhelper.js
```diff
@@ -0,0 +1,149 @@
...
+  // Functions
+  // ==================================================
+
+  // update strength level of password
+  function updateStrengthMeter(inputValue, meter, meterClass) {
+    if (typeof zxcvbn !== "undefined") {
+      meter.value = zxcvbn(inputValue).score;
+      Sage.meter.updateMeter(meterClass);
+    } else {
+      meter.closest(".sage-meter").style.opacity = "0";
```
teenwolfblitzer: Open to suggestions on a better way to treat this - currently I'm visibly hiding the meter when the external lib isn't loaded, but `display: none` may be better suited.

app/assets/javascripts/sage/system/inputhelper.js
```diff
@@ -0,0 +1,149 @@
+Sage.inputhelper = (function() {
+
+  // ==================================================
+  // Variables
+  // ==================================================
+  var spcChar = /(?=.*[~`!@#$%^&*|(){}/=:;.,<>+-])/g;
```
teenwolfblitzer: Used https://regexr.com and https://regex101.com/ for testing these regex strings, but could use 👀 on validity/improvement.

We'll also want to refactor these at some point to allow for varying the minimum number or required symbols/numbers/characters

app/assets/javascripts/sage/system/inputhelper.js
```diff
@@ -0,0 +1,149 @@
...
+  function bindPWEvents(helper) {
+    var fieldID = helper.getAttribute("data-js-helper-target"),
+        field = document.getElementById(fieldID);
+
+    var helperList = Sage.util.nodelistToArray(helper.querySelectorAll(".sage-hint__list-item")),
+        helperReqItems = helperList.map(function(ele) {
+          return {
+            id: ele.id,
+            type: ele.getAttribute("data-js-hint-type") || null,
+            minValue: ele.getAttribute("data-js-hint-min") || 0
```
monkeypox8, line 110: Can remove this. For future adjustment of the minimum value required.

philschanely (approved):
Works like a charm, @monkeypox8 ! Well done. Is it possible to improve the text contrast inside the password strength meter? It appears less than readable for most of the bar colors to me. Not sure if that can be controlled or not?

app/assets/stylesheets/sage/system/patterns/elements/_meter.scss
```diff
@@ -5,7 +5,9 @@
 ================================================== */
 
 $-sage-meter-height: rem(18px);
-$-sage-meter-color-bg: sage-color(primary, 100);
+$-sage-meter-color-bg: sage-color(white);
+$-sage-meter-color-label: sage-color(charcoal, 100);
+$-sage-meter-label-shadow: 1px 1px 0 rgba(255, 255, 255, 0.67);
```
AndrwM: @philschanely – I agree about the `.sage-meter__label` contrast.

This is black at ~70%ish opacity via `rgba(0,0,0,0.7)`.
Maybe we could increase the contrast like this?

screencast 2020-05-20 15-01-27

app/assets/javascripts/sage/docs/inputhelper.js
```diff
@@ -0,0 +1,20 @@
...
+    helperExample.classList.add('sage-input-helper--visible');
+    helperExample.style.position = 'relative';
+  }
+
+
+  return {
+    init: init
+  };
+
+})();
```
AndrwM, line 20: This is used in the design system docs and it forces the `.sage-input-helper` to be open for the preview?

I wonder if we should namespace scripts like this so they're differentiated from the other docs-related scripting? Not sure if that's overthinking this, we'll likely have more items like this.

We could possibly do something like `Sage.docs.previewUtil.inputhelper` or `Sage.docs.example.inputhelper`. Though, I don't know much about the docs scripts or the common patterns happening there.

app/assets/javascripts/sage/system/inputhelper.js
AndrwM: This seems good to me, could you make a comment here describing that this is a fallback in the case `zxcvbn` isn't loaded?

We'll have a better way to manage these dependancies once we're webpackin'.

app/assets/javascripts/sage/system/inputhelper.js
```diff
@@ -0,0 +1,148 @@
...
+    var criteria = document.getElementById(itemID);
+
+    if (bool === true ) {
+      criteria.classList.add(passingClass);
+      return true;
+    } else {
+      criteria.classList.remove(passingClass);
+      return false;
+    }
+  }
```
AndrwM: Nice!

app/assets/javascripts/sage/system/inputhelper.js
```diff
@@ -0,0 +1,148 @@
...
+            id: ele.id,
+            type: ele.getAttribute("data-js-hint-type") || null
+          }
+        });
+
+    var meter = helper.querySelector(".sage-meter__bar");
+
+    if (Sage.util.isIE()) {
+      focusBlurIE(field);
+    }
```
AndrwM, line 124: 😢 The things we do for you IE.

app/assets/javascripts/sage/system/inputhelper.js
AndrwM: Looks great, nice! I could see these becoming part of a global JS variables file later.

app/assets/javascripts/sage/system/inputhelper.js
AndrwM: Another note, in previewing this I noticed that the meter is given 100% success even when a symbol isn't included:

image

app/assets/stylesheets/sage/system/patterns/elements/_meter.scss
```diff
@@ -5,7 +5,9 @@
...
 $-sage-meter-bar-low: sage-color(red);
 $-sage-meter-bar-med: sage-color(orange);
 $-sage-meter-bar-optimum: sage-color(sage, 200);
```
AndrwM: What do you think about using `$color-success` or `$color-green-light` for this?

AndrwM (commented):
Really nice stuff. Can't wait to have this in Kajabi-Products. 👍 

Left a few comments. I think we shouldn't display a 100% password strength until the symbols criteria is qualified.

app/assets/javascripts/sage/system/inputhelper.js
teenwolfblitzer: Hmm, excellent point. Because the two items aren't coupled, I may have to add some sort of negative weighting value to the strength meter when all requirements haven't been fulfilled. The strength meter is meant more as a "guide" but ends up giving a false indication of a passing grade.

app/assets/stylesheets/sage/system/patterns/elements/_meter.scss
teenwolfblitzer: That does appear to be more readable. I may also lighten the tint of the background/meter color to increase the contrast some more.

app/assets/javascripts/sage/docs/inputhelper.js
teenwolfblitzer: Right, it's specifically for the preview. The docs scripts could use distinct filenames to help this as well, `inputhelper-example.js` or `inputhelper-docs.js` would clarify that it's not for public consumption. We don't have a lot of consistency as the docs structure has been evolving, so perhaps we'll need a follow-up issue cleaning up the documentation-only JS.

teenwolfblitzer:
@AndrwM Thanks for the feedback! Made some edits to address the issues you noted with scoring along with @philschanely 's notes on the label contrast.

app/views/sage/examples/objects/input_helper/_props.html.erb
```diff
@@ -0,0 +1,6 @@
+<tr>
```
cmcfadzean: Can probable delete the content here if not applicable at this time

app/views/sage/examples/objects/input_helper/_rules_dont.html.erb
```diff
@@ -0,0 +1,3 @@
+<ul>
```
cmcfadzean: Can probable delete the content here if not applicable at this time

app/assets/stylesheets/sage/system/patterns/elements/_meter.scss
```diff
@@ -5,11 +5,13 @@
...
 
 $-sage-meter-height: rem(18px);
-$-sage-meter-color-bg: sage-color(primary, 100);
-$-sage-meter-bar-low: sage-color(red);
-$-sage-meter-bar-med: sage-color(orange);
-$-sage-meter-bar-optimum: sage-color(sage, 200);
-$-sage-meter-bar-max: sage-color(sage);
+$-sage-meter-color-bg: sage-color(white);
+$-sage-meter-color-label: rgba(sage-color(black), 0.7);
+$-sage-meter-label-shadow: 1px 1px 0 rgba(255, 255, 255, 0.67);
```
cmcfadzean, line 10: In the screenshots, the label shadow is throwing me off. It seems out of alignment with the other look and feel. 
Curious how design feels about this. I am tempted to say that the label is not even needed and it could just be the color indicator bar and that being clear enough.

cmcfadzean (commented):
Just some minor thoughts.

app/assets/stylesheets/sage/system/patterns/elements/_meter.scss
teenwolfblitzer: Good call, I'll check with design. We don't have a full comp for this yet, so the bar label is only necessary for a11y purposes.

teenwolfblitzer:
Feedback from design was to update the label of the meter to indicate strength in real-time (examples attached), which also allows us to reduce the height of the bar and minimize its emphasis.

To prevent this PR from languishing further, I'm planning to merge this in for now and address the revisions in a follow-up.

|  Example 1   |  Example 2  |
|--------|--------|
|image|image (1)|

PR #263, opened 2020-05-15
Meter element

teenwolfblitzer:
## Description
Adds a styled `meter` element to indicate measurement of a fixed/known quantity. Note that there is no fallback here for IE browsers for now, since the intended use is strictly as a visual enhancement.

- ~Also adds a small modification to the `sage-input` markup, allowing `min` values in number inputs (used for this example)~ removed to prevent breaking changes in the main app
- Animation is Chrome/Safari only, Firefox does not support animating the meter value 😢 

meter

## Related

- #261


pixelflips (approved):
👍 +1 LGTM - Nice work!

PR #223, opened 2020-05-02
Webpacker (2/3): Migrate AssetPipeline Assets

AndrwM:
Separate PR for migrating AssetPipeline Assets to Webpack to keep the diffs clean.

**Points to `ajm-webpack2` branch.**

PR #275, opened 2020-06-04
Add doc for file structure

jacobthemyth:
This is just a markdown file I created to wrap my head around the file structure. If you think it should be in the repo, the base branch is your PR, but mostly I just wanted to ask questions in the context of the directory layout which is hard to do in a normal diff.

### DIFF
--- a/structure.md
+++ b/structure.md
@@ -38,9 +38,57 @@
 │       │   └── _styles.html.erb
 │       ├── examples
 │       │   ├── elements
+│       │   │   ├── _element.html.erb
+│       │   │   ├── _element_preview.html.erb
+│       │   │   ├── button
+│       │   │   ├── checkbox
+│       │   │   ├── danger_button
+│       │   │   ├── description
+│       │   │   ├── form_input
+│       │   │   ├── form_select
+│       │   │   ├── form_textarea
+│       │   │   ├── label
+│       │   │   ├── link_button
+│       │   │   ├── live_stream_wrapper
+│       │   │   ├── loader
+│       │   │   ├── radio
+│       │   │   ├── switch
+│       │   │   ├── table
+│       │   │   └── tooltip
 │       │   ├── objects
+│       │   │   ├── _object.html.erb
+│       │   │   ├── _object_preview.html.erb
+│       │   │   ├── alert
+│       │   │   ├── assistant
+│       │   │   ├── banner
+│       │   │   ├── billboard
+│       │   │   ├── card
+│       │   │   ├── content
+│       │   │   ├── form_section
+│       │   │   ├── live_active_mic
+│       │   │   ├── live_avatar
+│       │   │   ├── live_option_menu
+│       │   │   ├── live_profile_card
+│       │   │   ├── live_stream_control
+│       │   │   ├── live_stream_footer
+│       │   │   ├── live_stream_header
+│       │   │   ├── live_stream_video_grid
+│       │   │   ├── live_user_control
+│       │   │   ├── nav
+│       │   │   ├── page
+│       │   │   ├── page_heading
+│       │   │   ├── pagination
+│       │   │   ├── panel
+│       │   │   ├── sidebar
+│       │   │   └── tabs
 │       │   ├── shared
+│       │   │   ├── _heading.html.erb
+│       │   │   ├── _props.html.erb
+│       │   │   ├── _quick_links.html.erb
+│       │   │   └── _rules.html.erb
 │       │   └── utilities
+│       │       ├── _utility.html.erb
+│       │       └── media
 │       ├── layouts
 │       │   ├── application.html.erb
 │       │   └── breakout.html.erb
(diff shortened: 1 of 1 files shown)


structure.md
```diff
@@ -0,0 +1,129 @@
...
+├── Gemfile
+├── Gemfile.lock
+├── MIT-LICENSE
+├── README.md
+├── Rakefile
+├── app/
+│   ├── assets/
+│   │   ├── fonts/
+│   │   │   ├── inter/
+│   │   │   └── sage/
```
jacobthemyth, line 12: Are these fonts only used in the docs? Or should they be in `lib/sage-engine/app/assets/fonts`?

jacobthemyth: Are these fonts only used in the docs? Or should they be in `lib/sage-frontend`? I guess I'm not actually sure whether things like fonts and images should be in `/lib/sage-engine` or `/lib/sage-frontend`, I could imagine that we might want to access them from Ruby in views and stuff, but is that possible if they're just in a node module? I guess I don't actually know how webpacker works 😬.

structure.md
```diff
@@ -0,0 +1,129 @@
...
+│   │   ├── images/
+│   │   │   ├── avatar/
+│   │   │   ├── billboard/
+│   │   │   ├── card/
+│   │   │   └── svg/
+│   │   │       ├── active-microphone.svg
+│   │   │       ├── loader.svg
+│   │   │       └── spinner.svg
```
jacobthemyth: Ditto above, I would think these should be shared with `kajabi-products`?

structure.md
```diff
@@ -0,0 +1,129 @@
...
+│   │   │       ├── init.js
+│   │   │       ├── inputgroup.js
+│   │   │       ├── inputhelper.js
+│   │   │       ├── meter.js
+│   │   │       ├── overlay.js
+│   │   │       ├── select.js
+│   │   │       ├── sidebar.js
+│   │   │       ├── table.js
+│   │   │       ├── tooltip.js
+│   │   │       └── util.js
```
jacobthemyth: So I think you might have removed these in your PR, but are these components that should be in `lib/sage-engine/app/assets/javascripts`?

structure.md
```diff
@@ -0,0 +1,118 @@
...
+│   │   ├── lib/
+│   │   │   ├── sage/
+│   │   │   │   ├── engine.rb
+│   │   │   │   └── version.rb
+│   │   │   └── sage.rb
+│   │   ├── sage.gemspec
+│   │   └── test/
+│   │       ├── controllers/
+│   │       │   └── sage
+│   │       │       └── pages_controller_test.rb
```
jacobthemyth: I'm pretty sure this can be deleted since `/test` is testing PagesController.

structure.md
```diff
@@ -0,0 +1,100 @@
...
+│   │   │       ├── loader.svg
+│   │   │       └── spinner.svg
+│   │   ├── javascripts/
+│   │   │   ├── docs/
+│   │   │   │   ├── banner.js
+│   │   │   │   ├── example.js
+│   │   │   │   ├── inputhelper.js
+│   │   │   │   ├── live-option-menu.js
+│   │   │   │   └── meter.js
+│   │   │   └── sage_docs.js
```
jacobthemyth: I think you deleted these in your PR, but I would think they're still required right? Although maybe `sage_docs.js` could just be `application.js`.

structure.md
```diff
@@ -0,0 +1,100 @@
...
+│   │   └── stylesheets/
+│   │       ├── docs/
+│   │       ├── sage_docs.css.scss
+│   │       ├── sage_system.css.scss
+│   │       └── system/
```
jacobthemyth: Ditto above, can these be used from Rails views if they're in `lib/sage-frontend`?

jacobthemyth (commented):
@AndrwM I added a file structure doc just to help me wrap my head around how things are laid out, I'm not sure if this should be in the repo, but I left some comments for things I'm not really sure about. This is awesome! I didn't consider having a node module be a separate directory, this seems really clean.

structure.md
AndrwM: @jacobthemyth everything in `app/assets/` will be moved to `lib/sage-frontend`.

I haven't moved this yet because 90% of our PRs in this project are made to the `app/assets` directory. My plan here is to have a 24 hour pause on shipping PRs for our team and converting whatever is current in our master branch to `lib/sage-frontend`. This conversion won't take longer than an hour or two (I did it in a scrapped PR here: https://github.com/Kajabi/sage/pull/223).

Now that we do have a separation of "docs" and "system". I am also considering either:
1. Keeping the `sage_docs.css.scss` and its file dependancies within the "docs" Rails app assetpipeline.
2. Convert to webpack but keep it within the "docs" rails app.

structure.md
AndrwM: Great will do

structure.md
AndrwM: Yep! These would all be in `lib/sage-frontend`.

AndrwM:
@jacobthemyth I'll go ahead and move what is current in `app/assets` to `lib/sage-frontend` so its more representative of what will ship.

Sorry about the confusion here :)

structure.md
jacobthemyth: So I was kind of confused yesterday, I was thinking that if things like stylesheets or images were only exposed via a node module, we couldn't use things like `asset_path` in Rails, which I still think is true, but I think the thing I was missing is that you can access them with `image_pack_tag` which I forgot existed 😬.

Anyway, yeah this whole thing makes sense and is gonna be great 🤘. Closing this PR, but feel free to pull in the structure doc if it would be helpful.

jacobthemyth:
> @jacobthemyth I'll go ahead and move what is current in `app/assets` to `lib/sage-frontend` so its more representative of what will ship.
> 
> Sorry about the confusion here :)

That's no necessary I don't think, I pretty sure I understand now, so follow your heart on how this is laid out until the freeze.

PR #282, opened 2020-06-12
Deploy Docs [v0.17.1]

AndrwM:
- Fix breakout layout
- Fix favicon linking
- Implement Lockup into Docs site (match `sage_public` implementation)
- Display current version number
- Describe tag cutting workflow within ReadMe
- Add a version bump lifecycle hook for yarn to autodeploy the latest tag to master

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -24,3 +24,5 @@ group :development do
   # background. Read more: https://github.com/rails/spring
   gem 'spring'
 end
+
+gem "lockup", "~> 1.5"
--- a/README.md
+++ b/README.md
@@ -89,7 +89,7 @@ $ yarn version --minor
 **Note:** The Sage version is defined by the `./package.json` version and the version-tagged git commit. Please ensure these values match after a version bump.
 
 ### Update Kajabi-Products To The Latest Sage Version
-Our main app is uses a version-tagged commit from the Kajabi/Sage master branch as the source for the Sage frontend dependency.
+Our main app uses a version-tagged commit from the Kajabi/Sage master branch as the source for the Sage frontend dependency.
 
 This can be done using:
 ```bash
--- a/README.md
+++ b/README.md
@@ -70,9 +70,23 @@ It's recommended to run the Sage documentation site in tandem with Kajabi-Produc
 
 ## Deploying Sage
 ### Cut A New Version Of Sage
-The Sage version is set via the `./package.json` version.
+#### Git Remote Installation
+Before bumping the Sage version please ensure you have deploy rights to our [Heroku Sage-Design-System app](https://dashboard.heroku.com/apps/sage-design-system/) and you have Heroku app defined as a git remote location named `heroku`. For more information on how to set a Heroku app as a git remote see the [Heroku help docs](https://devcenter.heroku.com/articles/git#for-an-existing-heroku-app).
 
-It is recommended to use the [`yarn version`](https://classic.yarnpkg.com/en/docs/cli/version/) command for this because it also handles tagging the git commit.
+#### Create A Draft Release in Github
+Create a draft release tagged with the future version-tagged git commit (example: `v0.17.1`) that will be landing in `master` if one does not already exist. Describe the changes incoming since the last version, this is our official changelog.
+
+Save this release "As Draft" in Github. It will be automatically released when the version-tagged commit is merged into `master`.
+
+#### Version Bump & Deploy Docs With `yarn version`
+Use [`yarn version`](https://classic.yarnpkg.com/en/docs/cli/version/) to automatically bump your frontend package version with a version-tagged git commit. This command also will automatically deploy this version bump to the [sage-design-system.kajabi.com](https://sage-design-system.kajabi.com/) public documentation.
+
+```bash
+$ yarn version --minor
+# Use the --minor or --patch flags to update the minor or patch version number respectively
+```
+
+**Note:** The Sage version is defined by the `./package.json` version and the version-tagged git commit. Please ensure these values match after a version bump.
 
 ### Update Kajabi-Products To The Latest Sage Version
 Our main app is updated by pulling setting a version-tagged commit from the Kajabi/Sage master branch as the source for the Sage frontend dependency.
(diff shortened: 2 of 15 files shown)


pixelflips (approved):
👍 +1 - LGTM!

app/views/lockup/lockup/unlock.erb
```diff
@@ -0,0 +1,31 @@
+<div class="sage-container sage-container--xs" style="padding-top: 200px;">
```
AndrwM, line 1: I know this is _awful_ practice, however this is a one-off view and we don't have spacing classes within our codebase yet (do we have a consensus on if we want spacing classes in our codebase?)

Suggestions?

app/views/lockup/lockup/unlock.erb
```diff
@@ -0,0 +1,31 @@
...
+  <% if @wrong %>
+    <div class="sage-banner sage-banner--warning">
+      <p class="sage-banner__title">Hmm&hellip; that doesn&#8217;t seem right. Try again?</p>
+      <% if lockup_hint %>
+        <p class="sage-banner__copy"><%= lockup_hint %></p>
+      <% end %>
+    </div>
+  <% end %>
+
+  <%= form_for :lockup, url: { action: 'unlock' }, html: {style: "padding-top: 50px"} do |form| %>
```
AndrwM, line 16: Same thing here. These are the only 2 spots/

cmcfadzean (approved):
Looks great to me

README.md
```diff
@@ -70,12 +70,26 @@ It's recommended to run the Sage documentation site in tandem with Kajabi-Produc
...
+```bash
+$ yarn version --minor
+# Use the --minor or --patch flags to update the minor or patch version number respectively
+```
+
+**Note:** The Sage version is defined by the `./package.json` version and the version-tagged git commit. Please ensure these values match after a version bump.
 
 ### Update Kajabi-Products To The Latest Sage Version
-Our main app is updated by pulling setting a version-tagged commit from the Kajabi/Sage master branch as the source for the Sage frontend dependency.
+Our main app is uses a version-tagged commit from the Kajabi/Sage master branch as the source for the Sage frontend dependency.
```
jacobthemyth: Just a tiny typo 
```suggestion
Our main app uses a version-tagged commit from the Kajabi/Sage master branch as the source for the Sage frontend dependency.
```

Gemfile
```diff
@@ -24,3 +24,5 @@ group :development do
   # background. Read more: https://github.com/rails/spring
   gem 'spring'
 end
+
+gem "lockup", "~> 1.5"
```
jacobthemyth, line 28: I'd never seen lockup, super 🆒.

config/environments/production.rb
```diff
@@ -1,4 +1,7 @@
 Rails.application.configure do
+  ENV['LOCKUP_CODEWORD'] = "SageUX"
+  ENV['LOCKUP_HINT'] = "Password can be found in the GitHub repo README."
+
```
jacobthemyth: It's probably not a big deal, but if you ever wanted to change the codeword without committing and deploying, you can set this as a config var in Heroku instead: https://dashboard.heroku.com/apps/sage-design-system/settings

jacobthemyth (approved):
🤘

app/views/application/_sidebar.html.erb
```diff
@@ -22,5 +22,6 @@
         <li><%= link_to "Changelog", "https://github.com/Kajabi/sage/releases", class: "sage-nav__link", target: "_blank", rel: "noopener noreferrer" %></li>
       </ul>
     </nav>
+    <%= link_to "Version #{$SAGE_VERSION}", "https://github.com/Kajabi/sage/releases/tag/v#{$SAGE_VERSION}", class: "sage-btn sage-btn--small sage-btn--secondary", target: "_blank", rel: "noopener noreferrer" %>
```
teenwolfblitzer: Thanks for this! Been hoping we could add that in at some point 🙇 .

The link does break if the release is still in draft mode, but that shouldn't be a regular problem.

For a future follow-up, I'm wondering if the button could live in the header/`sage-assistant__actions` since we already have the "Changelog" link next to it.

app/views/lockup/lockup/unlock.erb
```diff
@@ -0,0 +1,31 @@
...
+    <% if params[:return_to].present? %>
+      <%= form.hidden_field "return_to", value: params[:return_to] %>
+    <% elsif @return_to.present? %>
+      <%= form.hidden_field "return_to", value: @return_to %>
+    <% end %>
+
+    <div class="sage-input-group sage-input-group--search" role="group">
```
teenwolfblitzer: This probably was left in by mistake; we shouldn't need the `sage-input-group--search` modifier here.

app/views/lockup/lockup/unlock.erb
teenwolfblitzer: I suppose we could create specific styles for this page/module in the docs. But since it's a one-time view that people shouldn't see more than a couple times a year (if that), I'd say we give it a pass for now as-is.

On the spacing class front: I can see the benefits for the ability to space items, but I'd argue against creating utility classes specific to those purposes—we'd likely end up fighting specificity across the app, and consistency between margin/padding between elements. But maybe we do introduce classes/patterns for frequent-used layouts. For example, centering content on a page like we have here.

teenwolfblitzer (approved):
👍 🔥 LGTM, one minor item to take a look at with an extra class in the input group, great work!

app/views/lockup/lockup/unlock.erb
AndrwM: @monkeypox8 Agreed. I feel similarly. 👍 

If I had a nickel for everytime I've seen someone "cheat" a design system with spacing classes. ;)

config/environments/production.rb
AndrwM: Updated 👍

app/views/application/_sidebar.html.erb
AndrwM: Great idea, updated 👏

PR #283, opened 2020-06-12
Introduce review apps

jacobthemyth:
## Description
This makes changes necessary to automatically create review apps on PRs.

Heroku requires a "real" database, so this adds the `pg` gem. I don't think the app even uses the database currently, but I figured it wouldn't hurt. All the devs will already need to have PostgreSQL installed to do app development.


AndrwM (approved):
+1 Thanks!

PR #285, opened 2020-06-18
Remove default line height [0.17.2]

philschanely:
## Description
Removes the root-level line-height setting that is no longer needed due to the setup for type specs added in prior releases.

Also makes small update to README to install database.


pixelflips (approved):
👍 +1 - LGTM!

PR #288, opened 2020-06-18
Deploy Failsafe + Better Docs

AndrwM:
## Description
- This prevents `yarn version` version bumping and git commit tagging from being successful if the deploy will fail afterward because the git remote location for production is missing
- Better deployment installation docs
- Ability to run a standalone deployment with `yarn deploy`

@philschanely this resolves the issues we were having.


philschanely (approved):
❤️

AndrwM:
@philschanely – Thanks! Shipping.

PR #289, opened 2020-06-19
Sage docs sidebar nav modifications 

teenwolfblitzer:
## Description
Slight modification of sidebar navigation output to improve link hierarchy. Also added a couple of stand-in styles for missing `:focus` and `:active` states and widened the sidebar on mobile devices.

## Related

- n/a

### DIFF
--- a/lib/sage-frontend/stylesheets/system/patterns/objects/_nav.scss
+++ b/lib/sage-frontend/stylesheets/system/patterns/objects/_nav.scss
@@ -13,6 +13,7 @@
 .sage-nav__link {
   display: flex;
   align-items: center;
+  margin-bottom: rem(2px);
   padding: sage-spacing(2xs) sage-spacing(xs);
   color: sage-color(charcoal, 400);
   border-radius: sage-border(radius);
--- a/lib/sage-frontend/stylesheets/system/patterns/objects/_nav.scss
+++ b/lib/sage-frontend/stylesheets/system/patterns/objects/_nav.scss
@@ -21,17 +21,22 @@
     @extend %t-sage-body-med;
   }
 
-  &:hover {
+  &:hover,
+  &:focus {
     background: sage-color(grey, 300);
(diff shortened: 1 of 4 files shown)


lib/sage-frontend/stylesheets/system/patterns/objects/_nav.scss
```diff
@@ -21,17 +21,22 @@
     @extend %t-sage-body-med;
```
AndrwM, line 22: What do you think about adding a 1px or 2px bottom margin to these navigation links (`.sage-nav__link`)?

image

This is more of an aesthetic change, it gives the rollover color block a little room to breathe.

AndrwM (commented):
I'm seeing this issue in the review app:

image

It may have something to do with a missing closing tag or a flex-grow issue with `.sage-content`. I can't tell from the browser.

teenwolfblitzer:
> I'm seeing this issue in the review app:
> 
> image
> 
> It may have something to do with a missing closing tag or a flex-grow issue with `.sage-content`. I can't tell from the browser.

whoa! I totally missed seeing that somehow. will take a look.

lib/sage-frontend/stylesheets/system/patterns/objects/_nav.scss
teenwolfblitzer: Nice! I like that, looks cleaner with the small spacing between links. I'll add that in.

I was hoping to circle back on the sidebar styling for a full refresh at some point. Chris had mocked up a version to use in the app, though I don't think it was final. Now that we're really close to landing Sage in the admin, hopefully one of us will be able to follow-up on that:

AndrwM: Sweet! I love this!

philschanely (approved):
🙌

Issue #287, opened 2020-06-18
Sage component needs for Build's Site Navigation rework


AndrwM:
# Sage Components Missing

Whiteboarded comps: https://www.figma.com/proto/XqyDa4KJg8qKy62HULoNyA/Website?node-id=2%3A8&viewport=-267%2C492%2C0.3764588236808777&scaling=min-zoom

- [x] Tabs
- [x] Button Dropdown
- [ ] "Orderable" Items
- [x] Modal

See below for screencaps

--------------------

## Tabs
![image](https://user-images.githubusercontent.com/565743/85058478-2dbc7380-b170-11ea-8649-72a59477811c.png)
![image](https://user-images.githubusercontent.com/565743/85058621-5e9ca880-b170-11ea-8123-af536d99a705.png)

-----------------

## Button Dropdown
![image](https://user-images.githubusercontent.com/565743/85058533-3f9e1680-b170-11ea-82aa-a3f748595201.png)
![image](https://user-images.githubusercontent.com/565743/85058922-cfdc5b80-b170-11ea-9836-8c6be0175bb4.png)

-------------------

## "Orderable" Items
![image](https://user-images.githubusercontent.com/565743/85058685-7411d280-b170-11ea-9776-ab82562516b4.png)
![image](https://user-images.githubusercontent.com/565743/85058726-81c75800-b170-11ea-87b3-eddccb6305d1.png)

PR #290, opened 2020-06-22
Sage Dropdown

AndrwM:
## Description
Forks functionality and style from `sage-live-option-menu`
Sage label modifier fix

### Screenshots
screencast 2020-06-23 17-50-47

## Related
#287


app/views/examples/objects/dropdown/_markup.html.erb
```diff
@@ -0,0 +1,16 @@
+<div class="sage-dropdown" role="dropdown" aria-label="Participant options">
```
teenwolfblitzer: I think we may need to alter the parent structure of this somewhat, as the menu inside of the button doesn't appear to be accessible via keyboard.

Usually, a button can be actuated using the space bar or enter key. But even if we used a `button` element here in place of the `.sage-btn` `div`, the embedded structure may seem confusing to a screenreader. We'd probably need to set/unset `tabindex` on the menu container as the button is expanded/contracted, for example. I'm also not sure if the button 'click' would also intercept keyboard navigation to a menu option.

I think what we'll end up needing is a container surrounding a button with the `.sage-dropdown` menu as a sibling. A good example of this can be seen here: https://www.w3.org/TR/wai-aria-practices/examples/menu-button/menu-button-links.html which is reaaally close to what you've got in place.

lib/sage-frontend/stylesheets/system/patterns/objects/_dropdown.scss
```diff
@@ -0,0 +1,123 @@
...
+================================================== */
+
+$-dropdown-height: rem(40px);
+$-dropdown-width: rem(224px);
+
+.sage-dropdown {
+  &::before {
+    content: "";
+    visibility: hidden;
+    z-index: (sage-z-index(modal) - 1);
```
teenwolfblitzer: `sage-z-index` can take incremental values, so you should be able to go with `sage-z-index(modal, -1)` here. I haven't tested with negative values, but it should work. Or actually, `sage-z-index(underlay)` might be even better

lib/sage-frontend/stylesheets/system/patterns/objects/_dropdown.scss
```diff
@@ -0,0 +1,123 @@
...
+    transition: 0.1s ease-in-out;
+    transition-property: border-bottom-color, left, right;
+  }
+
+  &:hover {
+    color: sage-color(charcoal, 500);
+    background-color: sage-color(grey, 200);
+  }
+
+  &:focus {
```
teenwolfblitzer, line 111: Could we also add an `:active` state here? Once we have keyboard navigation functional it'll be a helpful indicator. Possibly the same style as `:focus` with a different border color?

teenwolfblitzer (changes requested):
This is looking great, but I think we'll need to address the keyboard accessibility of the menu and button as noted below. It seems likely we'll have necessary functions embedded within the menus once rolled out to the app.

lib/sage-frontend/stylesheets/system/patterns/objects/_dropdown.scss
```diff
@@ -0,0 +1,151 @@
...
+    background-color: sage-color(primary, 300);
+    pointer-events: none;
+    opacity: 0.04;
+    cursor: default;
+
+    @include position-full;
+  }
+}
+
+[aria-expanded="true"] ~ .sage-dropdown__container {
```
philschanely: Can this be nested in `__container`?

lib/sage-frontend/stylesheets/system/patterns/objects/_dropdown.scss
```diff
@@ -0,0 +1,151 @@
...
+  visibility: visible;
+
+  &::before {
+    visibility: visible;
+    pointer-events: auto;
+  }
+}
+
+[aria-expanded="true"] ~ .sage-dropdown__container .sage-dropdown__menu {
```
philschanely: Can this be nested in `__menu`?

philschanely (approved):
Very nice! Some small suggestions but 👍 otherwise.

PR #291, opened 2020-06-22
adding voice and tone guides

cmcfadzean:
## Description
This branch adds some voice and tone guidance to our documentation as we look at improving the copy in our application this is just the first step.


philschanely (approved):
Has there been/should there be any institutional review of the message conveyed? I don't assume to judge it myself but from a code standpoint and principle, 👍

cmcfadzean:
Pulled this from a marketing guide that already exists. Also spoke with Orlando our CMO and he is comfortable with these rules. Thanks for checking in on that @philschanely great call.

PR #301, opened 2020-06-25
Prevent reboot from breaking within admin iOS buttons

AndrwM:


philschanely (approved):
No concerns at this time.

PR #307, opened 2020-06-26
Typography updates

teenwolfblitzer:
## Description
Various modifications affecting global typography settings. For the most part, naming will remain the same as before, and only the values should have changed.

tldr;
- We're now using a major second scale (a 9:8 ratio, or 1.125 multiplier), vs the previous minor third scale (6:5 ratio, or 1.2 multiplier)
- Headings `H1` - `H6` are smaller, less bold, and more tightly kerned overall
- `line-height` updates:
  - decreased from `1.875` to `1.75` on base text
  - heading `line-height` stays at `1.25`
- `letter-spacing` on base text decreased to `0.2px`
- `xbold` font weight option removed (this had only been used on headings, which are now `bold`/`700`)
- revised documentation:
  - standardized font sizing units to match Figma, with `px` instead of `pt`
  - replace `leading` with `line-height` to align with CSS notation
  - adjust vertical alignment in font spec tables

---

#### `sage-font-size`

| Spec  | New value          | Old value |
| ----- | ------------------ | --------- |
| `xs`  | `12.64px`          | `11.11px` |
| `sm`  | `14.22px`          | `13.33px` |
| `md`  | `16px` (NO CHANGE) | `16px`    |
| `lg`  | `18px`             | `19.2px`  |
| `xl`  | `20.25px`          | `23.04px` |
| `2xl` | `22.78px`          | `27.65px` |
| `3xl` | `25.63px`          | `33.18px` |
| `4xl` | `39.81px`          | `28.83px` |

---

#### `sage-font-weight`
| Spec    | New value | Old value |
| ------- | --------- | --------- |
| `xbold` | (REMOVED) | `800`     |

All other values remain the same

---

#### `sage-line-height`
| Spec | New value | Old value |
| ---- | --------- | --------- |
| `lg` | `1.75`    | `1.875`   |

All other values remain the same

---

#### `sage-letter-spacing`
| Spec | New value     | Old value |
| ---- | ------------- | --------- |
| `xs` | `0.2px` (NEW) | N/A       |

All other values remain the same

### Screenshots

|  NEW   |  OLD  |
|--------|--------|
|||

### Steps for testing
1. Visit the documentation site
2. Navigate through and verify that the typography adjustments have not unexpectedly affected alignment and spacing

### DIFF
--- a/app/views/pages/typography.html.erb
+++ b/app/views/pages/typography.html.erb
@@ -49,72 +49,72 @@
       <tr>
         <td><span class="t-sage-heading-1">Heading 1</span></td>
         <td><code>t-sage-heading-1</code>
-        <td>800</td>
-        <td>39.81pt</td>
+        <td>700</td>
+        <td>28.83pt</td>
         <td>1.25</td>
-        <td>0.8px</td>
+        <td>1px</td>
       </tr>
       <tr>
         <td><span class="t-sage-heading-2">Heading 2</span></td>
         <td><code>t-sage-heading-2</code>
-        <td>800</td>
-        <td>33.18pt</td>
+        <td>700</td>
+        <td>25.63pt</td>
         <td>1.25</td>
-        <td>0.8px</td>
+        <td>1px</td>
       </tr>
       <tr>
         <td><span class="t-sage-heading-3">Heading 3</span></td>
         <td><code>t-sage-heading-3</code>
         <td>700</td>
-        <td>27.65pt</td>
+        <td>22.78pt</td>
         <td>1.25</td>
-        <td>0.5px</td>
+        <td>1px</td>
       </tr>
       <tr>
         <td><span class="t-sage-heading-4">Heading 4</span></td>
         <td><code>t-sage-heading-4</code>
         <td>600</td>
-        <td>23.04pt</td>
+        <td>20.25pt</td>
         <td>1.25</td>
-        <td>0.5px</td>
+        <td>0.8px</td>
       </tr>
       <tr>
         <td><span class="t-sage-heading-5">Heading 5</span></td>
         <td><code>t-sage-heading-5</code>
         <td>600</td>
-        <td>19.2pt</td>
+        <td>18pt</td>
         <td>1.25</td>
-        <td>0.3px</td>
+        <td>0.8px</td>
       </tr>
       <tr>
         <td><span class="t-sage-heading-6">Heading 6</span></td>
         <td><code>t-sage-heading-6</code>
         <td>600</td>
         <td>16pt</td>
-        <td>1.875</td>
-        <td>0.3px</td>
+        <td>1.75</td>
+        <td>0.8px</td>
       </tr>
       <tr>
         <td><span class="t-sage-body">Body*</span></td>
         <td><code>t-sage-body</code>
         <td>400*</td>
         <td>16pt</td>
-        <td>1.875</td>
+        <td>1.75</td>
         <td>0.3px</td>
       </tr>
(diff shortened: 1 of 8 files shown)


teenwolfblitzer:
Noticed that the banner text is out of alignment while checking the documentation site, but it doesn't appear to be related to this change since it's also appearing on `master`.

I've opened up a new issue for a fix [here](https://github.com/Kajabi/sage/issues/308)

pixelflips (approved):
👍  - LGTM! I've listed a couple minor things below, but not entirely sure they're related to this update. No major show stoppers stood out. Nice work!

- **Banner**: text doesn't align with link or button.
- **Alert**: icon alignment slightly off.
- **Input group**: primary button positioning is slightly off

teenwolfblitzer:
> 👍 - LGTM! I've listed a couple minor things below, but not entirely sure they're related to this update. No major show stoppers stood out. Nice work!
> 
> * **Banner**: text doesn't align with link or button.
> * **Alert**: icon alignment slightly off.
> * **Input group**: primary button positioning is slightly off

@pixelflips Great catch on the Alert! I did notice the banner alignment just after posting this PR, and created a new issue up above. I'll have to do the same for the Alert since both bugs are also appearing in `master`. I'm not able to duplicate what you're seeing for the Input group, though. They look okay to me in Chrome/Firefox/Safari… could you post a screen?

pixelflips:
> @pixelflips Great catch on the Alert! I did notice the banner alignment just after posting this PR, and created a new issue up above. I'll have to do the same for the Alert since both bugs are also appearing in `master`. I'm not able to duplicate what you're seeing for the Input group, though. They look okay to me in Chrome/Firefox/Safari… could you post a screen?

@monkeypox8 Turns out it's only on the top level objects page. Sorry about that. I confused myself when I went back to double check. 🤦 

I also double checked on master and the issue is there as well, so just ignore me :)

Screen Shot 2020-06-26 at 3 28 41 PM

teenwolfblitzer:
> > @pixelflips Great catch on the Alert! I did notice the banner alignment just after posting this PR, and created a new issue up above. I'll have to do the same for the Alert since both bugs are also appearing in `master`. I'm not able to duplicate what you're seeing for the Input group, though. They look okay to me in Chrome/Firefox/Safari… could you post a screen?
> 
> @monkeypox8 Turns out it's only on the top level objects page. Sorry about that. I confused myself when I went back to double check. 🤦
> 
> I also double checked on master and the issue is there as well, so just ignore me :)
> 
> Screen Shot 2020-06-26 at 3 28 41 PM

Oh yeah, I see it now. Seems it's adding a margin on there, I may just add in a fix now.

EDIT: Input group addressed in #310 

Banner and Alert updates will be resolved in #308 and #309, respectively

app/views/pages/typography.html.erb
```diff
@@ -20,7 +20,7 @@
 
 <div class="sage-type">
   <p>
-    The Sage type system uses a minor third modular scale (factor of 1.2) with a base font size of 16pt.
+    The Sage type system uses a minor third modular scale (factor of 1.125) with a base font size of 16px.
```
philschanely: Per Chris's comment in Slack I believe that `minor third` should change to `major second`.

philschanely (approved):
👍 with one non-blocking comment.

app/views/pages/typography.html.erb
teenwolfblitzer: Oooh, nice catch! I'll update that part 👍

AndrwM:
LGTM +1!

PR #309, opened 2020-06-27
Reduce `line-height` in Alert title

teenwolfblitzer:
## Description
Per feedback from design, the `line-height` of the title in the Alert object should be decreased from its current `1.875` value to `1.5`. When the title text wraps, the vertical spacing appears too tall.

This also corrects a misalignment with the title and icon as seen in the attached screens.

|  before   |  after  |
|--------|--------|
|alert-before|alert-after|

## Related

- n/a


pixelflips (approved):
👍 +1

PR #316, opened 2020-06-30
Sortable, Button Icon-Only, Panel – for BUILD Site Navigation Project

AndrwM:
#### New
- Adds Sage Sortable object
- Sage Sortable scripting using SortableJS
- Add Sage JS util helper for sending an AJAX and receiving a JS response back, this is a common pattern in KJB-Products
- Sage Buttons now support a `sage-btn--icon-only-iconnamehere` modifier
  - This matches our syntax for other icon buttons (`sage-btn--icon-iconstyle-iconnamehere`)

#### Refactor
- Sage JS Init: Cleaner handling of checking whether the context is Sage Docs
- Sage Buttons style treatment update: Tertiary is now the new Secondary
  - Colab with: @philschanely, @croswell @cameronsimony 
- Sage Panel refactor to allow for headers, footers, and "header actions"
  - **Note:** Sage Panel syntax has now diverged from our usage in Docs
  - **#323, follow up PR has the fix for the Docs to keep this PR's diff scoped**

#### Bugs
- Fixes Sage Dropdown alignment issue and removes color overlay (ty @philschanely)
- Fixes Sage Tooltip caret alignment

## Screenshots
image
image
image

### DIFF
--- a/app/views/examples/objects/sortable/_markup.html.erb
+++ b/app/views/examples/objects/sortable/_markup.html.erb
@@ -1,0 +1,44 @@
+<div class="sage-sortable" data-js-sortable>
+
+  <div class="sage-sortable__item">
+    <div class="sage-sortable__item-content">
+      <h5 class="t-sage-body">Train Conducting</h5>
+      <p class="t-sage-body-xsmall">Assignment</p>
+    </div>
+    <div class="sage-sortable__item-actions">
+      <a class="sage-btn sage-btn--tertiary">Edit<a>
+    </div>
+  </div>
+
+  <div class="sage-sortable__item">
+    <div class="sage-sortable__item-content">
+      <h5 class="t-sage-body">Train Conducting</h5>
+      <p class="t-sage-body-xsmall">Assignment</p>
+    </div>
+    <div class="sage-sortable__item-actions">
+      <a class="sage-btn sage-btn--tertiary">Edit<a>
+    </div>
+  </div>
+
+  <div class="sage-sortable__item">
+    <div class="sage-sortable__item-content">
+      <h5 class="t-sage-body">Train Conducting</h5>
+      <p class="t-sage-body-xsmall">Assignment</p>
+    </div>
+    <div class="sage-sortable__item-actions">
+      <a class="sage-btn sage-btn--tertiary sage-btn--icon-preview-on"><a>
+      <a class="sage-btn sage-btn--tertiary sage-btn--icon-pen"><a>
+    </div>
+  </div>
+
+  <div class="sage-sortable__item">
+    <div class="sage-sortable__item-content">
+      <h5 class="t-sage-body">Train Conducting</h5>
+      <p class="t-sage-body-xsmall">Assignment</p>
+    </div>
+    <div class="sage-sortable__item-actions">
+      <a class="sage-btn sage-btn--tertiary">Edit<a>
+    </div>
+  </div>
+
+</div>
--- a/app/views/examples/objects/sortable/_markup.html.erb
+++ b/app/views/examples/objects/sortable/_markup.html.erb
@@ -6,7 +6,8 @@
       <p class="t-sage-body-xsmall">Assignment</p>
     </div>
     <div class="sage-sortable__item-actions">
-      <a class="sage-btn sage-btn--tertiary">Edit<a>
+      <a class="sage-btn sage-btn--tertiary sage-btn--icon-only-preview-on" href="#" data-js-tooltip="Preview">Preview</a>
+      <a class="sage-btn sage-btn--tertiary sage-btn--icon-only-pen" href="#" data-js-tooltip="Edit">Edit</a>
     </div>
   </div>
 
@@ -16,7 +17,8 @@
       <p class="t-sage-body-xsmall">Assignment</p>
     </div>
     <div class="sage-sortable__item-actions">
-      <a class="sage-btn sage-btn--tertiary">Edit<a>
+      <a class="sage-btn sage-btn--tertiary sage-btn--icon-only-preview-on" href="#" data-js-tooltip="Preview">Preview</a>
+      <a class="sage-btn sage-btn--tertiary sage-btn--icon-only-pen" href="#" data-js-tooltip="Edit">Edit</a>
     </div>
   </div>
 
@@ -26,8 +28,8 @@
       <p class="t-sage-body-xsmall">Assignment</p>
     </div>
     <div class="sage-sortable__item-actions">
-      <a class="sage-btn sage-btn--tertiary sage-btn--icon-preview-on"><a>
-      <a class="sage-btn sage-btn--tertiary sage-btn--icon-pen"><a>
+      <a class="sage-btn sage-btn--tertiary sage-btn--icon-only-preview-on" href="#" data-js-tooltip="Preview">Preview</a>
+      <a class="sage-btn sage-btn--tertiary sage-btn--icon-only-pen" href="#" data-js-tooltip="Edit">Edit</a>
     </div>
   </div>
 
@@ -37,7 +39,8 @@
       <p class="t-sage-body-xsmall">Assignment</p>
     </div>
     <div class="sage-sortable__item-actions">
-      <a class="sage-btn sage-btn--tertiary">Edit<a>
+      <a class="sage-btn sage-btn--tertiary sage-btn--icon-only-preview-on" href="#" data-js-tooltip="Preview">Preview</a>
+      <a class="sage-btn sage-btn--tertiary sage-btn--icon-only-pen" href="#" data-js-tooltip="Edit">Edit</a>
     </div>
   </div>
 
--- a/app/views/examples/objects/sortable/_markup.html.erb
+++ b/app/views/examples/objects/sortable/_markup.html.erb
@@ -1,47 +1,25 @@
-<div class="sage-sortable" data-js-sortable>
+<div class="sage-sortable" data-js-sortable="resourceNameHere">
 
-  <div class="sage-sortable__item">
+  <section class="sage-sortable__item" data-js-sortable-update-url="/update/url/here">
     <div class="sage-sortable__item-content">
-      <h5 class="t-sage-body">Train Conducting</h5>
-      <p class="t-sage-body-xsmall">Assignment</p>
+      <h1 class="t-sage-body">Batman</h1>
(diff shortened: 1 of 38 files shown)


pixelflips (commented):
@AndrwM I just pulled down this branch and it seems like the panels are missing padding. 

May not be specifically related to this branch, but I ran bundle, install, and relaunched my server and still seeing it. Just let me know.

AndrwM:
@pixelflips Yep! That's handled in a follow up PR here #323  to keep this diff clean

pixelflips:
> @AndrwM I just pulled down this branch and it seems like the panels are missing padding.
> 
> May not be specifically related to this branch, but I ran bundle, install, and relaunched my server and still seeing it. Just let me know.

@AndrwM nevermind I just saw your note about this in the description.

pixelflips (approved):
👍  LGTM! Sorry about the confusion around panels. Great work!!

app/views/examples/objects/sortable/_markup.html.erb
```diff
@@ -0,0 +1,25 @@
+<div class="sage-sortable" data-js-sortable="resourceNameHere">
+
+  <section class="sage-sortable__item" data-js-sortable-update-url="/update/url/here">
+    <div class="sage-sortable__item-content">
+      <h1 class="t-sage-body">Batman</h1>
```
philschanely, line 5: If these are type elements that will typically be a part of this component I wouldn't argue with making them elements of the block such as `sage-sortable__item-title` and sage-sortable__item-tagline` and moving the type spec setting therein.

lib/sage-frontend/javascript/system/init.js
```diff
@@ -1,8 +1,12 @@
 Sage.init = function(elementNamesToInit) {
-  var shouldInit = function(elementName, selector) {
+  let shouldInit = function(elementName, selector) {
```
philschanely, line 2: If we're OK moving to `let` etc. can we now safely also use arrow functions and imports too? Just curious.

lib/sage-frontend/javascript/system/sortable.js
```diff
@@ -0,0 +1,50 @@
+import Sortable from 'sortablejs/modular/sortable.core.esm.js';
+
+Sage.sortable = (function() {
+  // ==================================================
+  // Variables
+  // ==================================================
+  const SELECTOR_CONTAINER = 'data-js-sortable';
+  const SELECTOR_ITEM_UPDATE_URL = 'data-js-sortable-update-url';
+  const SETTINGS = {
+    ghostClass: 'sage-sortable__item--ghost',
```
philschanely, line 11: 👻

lib/sage-frontend/stylesheets/system/core/_variables.scss
```diff
@@ -85,7 +85,7 @@ $sage-btn-shadow-base: sage-shadow(sm) !default;
 $sage-btn-shadow-hover: sage-shadow(md) !default;
 
 // Focus outline
-$sage-btn-focus-outline-spacing: sage-spacing(2xs) !default;
+$sage-btn-focus-outline-spacing: ( sage-spacing(2xs) / 2) !default;
```
philschanely: Not just `rem(2px)`? :-P that or perhaps add `3xs`?

lib/sage-frontend/stylesheets/system/patterns/objects/_dropdown.scss
```diff
@@ -41,8 +41,7 @@ $-dropdown-item-focus-line-spacing-x: sage-spacing(sm);
 .sage-dropdown__menu {
   visibility: hidden;
   position: absolute;
-  left: 0;
-  top: calc(100% + #{sage-spacing(xs)});
+  top: calc(100% - #{( sage-spacing(xs) / 2)});
```
philschanely: Are you doing this math particularly to remind us that we are halving the `xs` value rather than just using `2xs`?

philschanely (approved):
A few small suggested fixes/questions but nothing blocking per se.

lib/sage-frontend/javascript/system/init.js
```diff
@@ -1,8 +1,12 @@
...
     return elementNamesToInit.includes(elementName) && document.querySelector(selector) !== null;
   };
 
+  let inDocumentationContext = function() {
+    return !!document.querySelector('.sage-docs');
```
teenwolfblitzer, line 7: 💥 💥

lib/sage-frontend/javascript/system/util.js
```diff
@@ -19,11 +19,29 @@ Sage.util = (function(Sage) {
...
+  function ajaxRequestWithJsResponse(urlTarget, urlParams) {
+    let xhr = new XMLHttpRequest();
+    xhr.open('POST', urlTarget);
+    xhr.setRequestHeader("Accept", "text/javascript");
+    xhr.setRequestHeader('X-CSRF-Token', document.querySelector('meta[name="csrf-token"]').content);
+    xhr.setRequestHeader('Content-Type', 'application/x-www-form-urlencoded; charset=UTF-8');
+
+    xhr.onload = function (evt) {
+      let elScript = document.createElement('script');
+      elScript.setAttribute('type', 'text/javascript')
```
teenwolfblitzer: not strictly necessary but missing a semicolon here

lib/sage-frontend/stylesheets/system/patterns/objects/_dropdown.scss
AndrwM: Haha, whoops – thanks for catching

lib/sage-frontend/stylesheets/system/core/_variables.scss
AndrwM: `3xs` is a great idea, thanks!

lib/sage-frontend/stylesheets/system/patterns/objects/_panel.scss
```diff
@@ -7,27 +7,35 @@
...
 .sage-panel__header {
-  margin-bottom: $sage-panel-padding;
+  padding: sage-spacing(xs) sage-spacing(xs) 0 $sage-panel-padding;
+  display: flex;
+  align-items: center;
+  border-bottom: sage-border();
+  min-height: rem(60px);
+}
+
+.sage-panel__header-right {
```
teenwolfblitzer, line 23: If this is a modifier, should it be `.sage-panel-header--right`?

lib/sage-frontend/javascript/system/init.js
AndrwM: You bet we can! 🎉 🎉 🎉

lib/sage-frontend/stylesheets/system/patterns/objects/_panel.scss
```diff
@@ -7,27 +7,35 @@
...
+  margin: 0 0 0 auto;
```
teenwolfblitzer, line 24: Since `.sage-panel__header` doesn't have a margin being set, it's likely this could be reduced to `margin-left: auto` so we're not resetting the other values (couldn't find an example in use)

app/views/examples/objects/sortable/_markup.html.erb
AndrwM: @croswell and I talked about this a bit. – My preference would be to stick with using these type helpers because I'm thinking in the future we can use different types of elements here (small headings, labels, etc). Using `sage-sortable__item-title` seemed like a pre-emptive optimization.

This is a pattern that is used else where for Product sorting so I wanted to keep it as flexible as possible.

lib/sage-frontend/stylesheets/system/patterns/objects/_sortable.scss
```diff
@@ -0,0 +1,73 @@
...
+}
+
+.sage-sortable__item-content {
+  z-index: sage-z-index();
+  width: 100%;
+  list-style: none;
+  overflow: hidden;
+
+  & > * {
+    user-select: none;
```
teenwolfblitzer: Curious if this was a stylistic decision to prevent the children from 'highlighting' when the parent is dragged. There _shouldn't_ be an accessibility hit, though it's probably best to err on the side of leaving this out if possible.

teenwolfblitzer (approved):
👍 🔥  LGTM, great work! A few non-blocking items, nice to see ES6 syntax up and running.

This is beyond the scope of this PR, but we'll want to address keyboard controls as part of the accessibility discussion: e.g. should the sortable items be keyboard selectable, and to what extent can that be made possible? My guess is it'd be a pretty heavy lift.

lib/sage-frontend/javascript/system/init.js
teenwolfblitzer: haha! that GIF is amazing

lib/sage-frontend/stylesheets/system/patterns/objects/_sortable.scss
AndrwM: Yep, a stylistic decision to prevent any accidental text selection on drag.
I'll remove 👍

lib/sage-frontend/stylesheets/system/patterns/objects/_panel.scss
AndrwM: This is a child of `.sage-panel__header`. – Just for right-aligned content in the header (buttons, label, small-text, etc).

Example I have in mind:
```
.sage-panel
  .sage-panel__header
    .t-heading-2
    .t-body
    .sage-panel__header-right // Elements in this container are "floated" next to each other
        .sage-btn
        .sage-label

```

I don't think I'm _all the way there_ as I haven't tested the configuration above what that's what I had in mind in terms of structure.

lib/sage-frontend/stylesheets/system/patterns/objects/_panel.scss
AndrwM: See above comment. This is a child of `.sage-panel__header`.

lib/sage-frontend/javascript/system/util.js
AndrwM: Ty!

AndrwM:
Fixed the issues found above. Landing. Thank you everyone!

lib/sage-frontend/stylesheets/system/patterns/objects/_panel.scss
teenwolfblitzer: Ah, I see. Disregard!

PR #317, opened 2020-07-01
Sage Breadcrumbs

philschanely:
## Description

Adds a formal Breadcrumbs element and revises the Page Heading object.

#### Basic setup

Screen Shot 2020-07-01 at 5 45 06 PM

#### Page Header Revised

Screen Shot 2020-07-01 at 5 45 25 PM


lib/sage-frontend/stylesheets/system/index.scss
```diff
@@ -45,37 +45,38 @@
 @import "utilities/hidden";
 
 // Elements
-@import "patterns/elements/meter";
-@import "patterns/elements/description";
-@import "patterns/elements/label";
-@import "patterns/elements/tooltip";
+@import "patterns/elements/breadcrumbs";
```
philschanely, line 48: Couldn't help myself here. Just. Had. To. Organize. 😂

app/views/examples/elements/breadcrumbs/_markup.html.erb
```diff
@@ -0,0 +1,15 @@
+<p class="sage-breadcrumbs<%= klass != "" ? " #{klass}" : "" %>">
+  <% if items %>
+    <% items.each_with_index do |item, i| %>
+      <% if i > 0 %>
+        <span class="sage-breadcrumbs__separator">/</span>
```
pixelflips: I'd be interested to hear what others think, but how would you feel about possibly using `::before` along with `content: "/"` to create the separator instead of the `span`?

app/views/examples/elements/breadcrumbs/_markup.html.erb
```diff
@@ -0,0 +1,15 @@
+<p class="sage-breadcrumbs<%= klass != "" ? " #{klass}" : "" %>">
```
pixelflips: Not a show stopper at all, but what do you think about breadcrumbs being marked up as an `ol` vs. using a `p`?

app/views/examples/elements/breadcrumbs/_markup.html.erb
philschanely: I love this idea... I was trying to keep the separator out of the hyperlink hit area by making it a separate thing. I do think I should add aria hidden though. Any pushback?

app/views/examples/elements/breadcrumbs/_markup.html.erb
philschanely: Love this too! 🙌 I do typically prefer lists for things like this. However it adds some complexity where this simpler solution may still read well. On this note perhaps the separators could have an alternative aria label that indicates the path from one page to the next and with us add more purpose to this as a paragraph. Thoughts?

app/views/examples/elements/breadcrumbs/_markup.html.erb
pixelflips: no pushback here, but if we wanted to keep the separator out of the link it might be worth using the `ol` and then we could attach it to the `li` instead.

app/views/examples/elements/breadcrumbs/_markup.html.erb
pixelflips: I was just checking out https://www.w3.org/TR/wai-aria-practices/examples/breadcrumb/index.html and thought I'd bring some of those things up, but definitely not wanting to cause any unwanted complexity.

pixelflips (approved):

app/views/examples/elements/breadcrumbs/_markup.html.erb
philschanely: Good call! I will rework to follow this model!

PR #320, opened 2020-07-07
Extend grid classes

teenwolfblitzer:
## Description
Grew tired of seeing the 'don't extend' warnings from Stylelint. The change here adds placeholders into the grid class generator. I'm still on the fence about this implementation, since generally we shouldn't be extending classes for layout. In either case, these do **not** get output in the compiled CSS.

Also adding a standalone `yarn lint` script in cases when spinning up the server isn't needed.

### Screenshots
No visual changes should be introduced here.

### Steps for testing
1. Visit the documentation site
2. Navigate to "Form section" under "Objects"
3. Verify the object looks and functions as expected with no changes.

## Related

- n/a

### DIFF
--- a/lib/sage-frontend/stylesheets/system/layout/_grid.scss
+++ b/lib/sage-frontend/stylesheets/system/layout/_grid.scss
@@ -54,6 +54,10 @@
 }
 
 @for $sage-i from 1 through $sage-grid-columns {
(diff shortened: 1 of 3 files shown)


teenwolfblitzer:
Closing this pending outcome of Sage grid/layout discussions

PR #323, opened 2020-07-08
Docs Container

AndrwM:
## Follow up to #316
`sage-panel` has diverged from our usage in the docs. Since it's only a basic color-boxed wrapper in the context of the docs it has been migrated and renamed to `docs-container`. A general "box" that can be used to wrap things.

### DIFF
--- a/app/views/application/_assistant.html.erb
+++ b/app/views/application/_assistant.html.erb
@@ -11,7 +11,7 @@
   <div class="sage-assistant__actions">
     <%= link_to "v#{$SAGE_VERSION}",
       "https://github.com/Kajabi/sage/releases/tag/v#{$SAGE_VERSION}",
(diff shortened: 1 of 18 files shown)


pixelflips (approved):
👍  +1

philschanely (approved):

teenwolfblitzer (approved):
👍 Should Panel/`sage-panel` now be removed from from the list of objects if it's no longer intended for consumption?

AndrwM:
@monkeypox8 This is in addition to `sage-panel`, the issue is we have about 40 instances of the old implementation. Since our docs styles are still flexible and we're assigning more constraints (and children... header, body, footer) to `sage-panel`, I duped the old `sage-panel` as `docs-container`.

PR #267, opened 2020-07-09
Revisions to input hint meter and label

teenwolfblitzer:
## Description
Drawing on examples from design provided below, we'll make better use of the label as an indicator of strength, and shorten the height of the meter itself to de-emphasize its appearance until a user has interacted with the associated field.

Chrome recently updated their meter styles, which has affected the current implementation (as seen in "Before" screenshot)

- [x] Remove Chrome `meter` border
- [x] Add dynamic descriptive text describing password strength

|  Before   |  After (default)  | After (filled)|
|--------|--------|--------|
|input-helper-old|input-helper-default|input-helper-filled|

### Inspiration
|  Example 1   |  Example 2  |
|--------|--------|
|image|image (1)|

## Related

- #261

### DIFF
--- a/lib/sage-frontend/javascript/system/inputhelper.js
+++ b/lib/sage-frontend/javascript/system/inputhelper.js
@@ -12,6 +12,14 @@ Sage.inputhelper = (function() {
   var passingClass = "sage-hint__list-item--success";
   var inputErrorClass = "sage-input--error";
   var sageMeterBar = ".sage-meter__bar";
+  var sageMeterBarDesc = document.querySelector(".sage-meter__description");
+
+  var meterLevels = {
+    low: 0.25,
+    med: 0.5,
+    high: 0.75,
+    max: 0.9
+  };
 
 
   // ==================================================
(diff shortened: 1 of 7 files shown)


lib/sage-frontend/stylesheets/system/patterns/objects/_input_helper.scss
```diff
@@ -29,28 +29,33 @@ $-sage-helper-decoration-list-success: line-through;
...
   opacity: 0;
 
-  &::before {
+  &::before,
+  &::after {
     content: "";
     display: block;
     position: absolute;
-    top: -#{$-sage-helper-pointer-size};
+    top: calc(-#{$-helper-pointer-size} - #{rem(1px)});
```
philschanely: I wonder if `calc()` is needed here?

lib/sage-frontend/javascript/system/inputhelper.js
```diff
@@ -25,17 +33,51 @@ Sage.inputhelper = (function() {
...
       meter.closest(".sage-meter").style.display = "none";
     }
   }
 
+  function limitScore(num) {
+    return num > 1 ? num - 1 : num;
+  }
+
+  // update strength level text description
+  function updateStrengthText(ele, score) {
```
philschanely, line 53: 😂 I've not seen `ele` as the abbreviation for `element` in a variable before.

philschanely (approved):
Nice work!! A few non-blocking notes

lib/sage-frontend/javascript/system/inputhelper.js
teenwolfblitzer: You know, I never really thought that much about it. Likely a habit inherited from using `evt` for `event`. I suppose those 4 characters don't really make much of a difference in bytes saved in any case 😀

lib/sage-frontend/stylesheets/system/patterns/objects/_input_helper.scss
teenwolfblitzer: Great catch! removing this one

lib/sage-frontend/javascript/system/inputhelper.js
philschanely: haha no worries! I suppose I'm accustomed to just `e` for event and `el` for elements. 🤷

PR #324, opened 2020-07-09
Add input prefix and suffix

philschanely:
## Description
Modifies labels and adds prefix and suffix to input fields

#### Label changes

|  before  |  after  |
|--------|--------|
| Screen Shot 2020-07-09 at 1 25 38 PM | Screen Shot 2020-07-09 at 1 23 34 PM |

#### New prefix/suffix

Screen Shot 2020-07-09 at 1 18 13 PM

### Testing

See Label Form Input Elements for changes.


lib/sage-frontend/stylesheets/system/patterns/elements/_label.scss
```diff
@@ -5,60 +5,82 @@
...
-}
-.sage-label--danger {
-  color: $sage-label-danger-text-color;
-  background-color: $sage-label-danger-color;
-  border-color: $sage-label-danger-color;
-  &.sage-label--subtle {
-    color: $-sage-label-subtle-danger-color;
-    background-color: sage-color(white);
-    border-color: $-sage-label-subtle-danger-border;
+@each $-color-name, $-color-settings in $-label-colors {
```
teenwolfblitzer, line 72: 😍

teenwolfblitzer (approved):
👍 🔥 LGTM! This may be an edge case, but I'm curious about your thoughts on if a user might click/focus on the prefix text, expecting it to act the same as an input label (placing focus in the field).

philschanely:
> 👍 🔥 LGTM! This may be an edge case, but I'm curious about your thoughts on if a user might click/focus on the prefix text, expecting it to act the same as an input label (placing focus in the field).

Great point. I'll definitely adjust accordingly!

PR #327, opened 2020-07-10
Add tiny button

philschanely:
## Description
Adds a "tiny" button variation. Intended mostly for situations with locked text like the hashtag and file examples shown below, and may be most relevant for the secondary and tertiary styling, but all possibilities are demonstrated here for posterity.


pixelflips (approved):
👍  +1

PR #328, opened 2020-07-10
Add sandbox and lint

philschanely:
## Description
Adds a sandbox page at /pages/sandbox for tinkering with stuff in context.

Also catches up on linting 🤓


pixelflips (approved):
👍  - I dig the ❤️   in the sandbox description. Thanks for handling the linting as well!

PR #331, opened 2020-07-13
Kajabi Admin Site Navigation: New JS Init Strategy, Sage Tabs, & Updates

AndrwM:
## Minor
- **New JS Init Strategy**
  - A "presence" listener that inits or unbinds a _single instance_ depending on whether it exists in the DOM
  - Event Listener that maps to a Sage JS command. Products can interact with Sage scripting by dispatching a JS event.
  - Converted the following components to our new JS init strategy:
      - Modal
      - Tooltip
      - Dropdown
      - Sortable
      - Tabs (These are new!)
  - Uses [ArriveJS](https://github.com/uzairfarooq/arrive) (same as on Products)
- **Sage Tabs**
  - Scripting to show/hide a panel
  - Radio button style variant
  - `.sage-tabs-pane` class that handles hiding and showing a tab pane (this bucks our organization conventions a bit)
- Sage Panel
  - Rename `header-right` to a more semantic `header-aside`
  -  New Sage Container Size (`modal`)
----
## Patch
- Sage Loader
  - Give it some default spacing
- Sage Dropdown
  - Put the dropdowns icons to the right of the option text
  - Small reorganization to make the file easier to read
- Sage Modal
  - Add animation
- Reboot: Remove dialog default browser styles


app/views/examples/objects/tabs/_markup.html.erb
```diff
@@ -1,5 +1,17 @@
...
+  <% options.each do |option| %>
+    <button type="button" href="#" class="sage-tabs__tab <%= option[:is_active] ? "sage-tabs__tab--active" : "" %>"><%= option[:text] %></button>
+  <% end %>
+</div>
+
+<div class="sage-tabs sage-tabs--style-radio">
+  <% options.each do |option| %>
+    <button type="button" href="#" class="sage-tabs__tab <%= option[:is_active] ? "sage-tabs__tab--active" : "" %>"><%= option[:text] %></button>
+  <% end %>
+</div>
```
AndrwM, line 17: Add Pane exampe

lib/sage-frontend/javascript/system/tooltip.js
```diff
@@ -1,55 +1,53 @@
...
+    tooltip.position = evt.target.getAttribute(`${DATA_ATTR}-position`);
     tooltip.dataItems = [`${DATA_ATTR}-theme`, `${DATA_ATTR}-size`];
 
     if (!tooltip.innerHTML.length > 0) return;
-    generateClasses(tooltip, e);
+    generateClasses(tooltip, evt);
 
     document.body.appendChild(tooltip);
-    positionTooltip(e.target, tooltip, pos);
+    positionTooltip(evt.target, tooltip, pos);
```
AndrwM, line 40: I did some cleanup here to clarify what datatype `e` represented.

lib/sage-frontend/stylesheets/system/patterns/objects/_modal.scss
```diff
@@ -33,38 +33,63 @@ $-modal-padding-y: sage-spacing(md);
...
+  margin: sage-spacing(lg) $-modal-padding-x $-modal-padding-y;
+
+  > :last-child {
+    margin-bottom: 0;
+  }
 }
 
 .sage-modal__content {
-  padding: $-modal-padding-y $-modal-padding-x;
+  margin: $-modal-padding-y $-modal-padding-x;
```
AndrwM, line 88: Using margin here to take advantage of margin collapsing.

philschanely:
@AndrwM reviewing now; consider running the linter as well. ❤️

app/views/examples/objects/modal/_markup.html.erb
```diff
@@ -1,14 +1,14 @@
-<div class="sage-modal" data-js-modal="cool-modal" role="dialog" aria-modal="true">
-  <div class="sage-modal__container">
-    <div class="sage-modal__header">
-      <h5 class="t-sage-heading-5">Example Sage Modal</h5>
-    </div>
+<dialog class="sage-modal" data-js-modal="cool-modal" role="dialog" aria-modal="true">
+  <section class="sage-modal__container">
+    <header class="sage-modal__header">
+      <h1 class="t-sage-heading-5">Example Sage Modal</h1>
```
philschanely: I'm seeing heading-4 as the type spec prescribed in the design.

lib/sage-frontend/stylesheets/system/patterns/objects/_modal.scss
```diff
@@ -33,38 +33,63 @@ $-modal-padding-y: sage-spacing(md);
...
+  }
+
+  @include _animation-off;
   z-index: sage-z-index(modal);
-  transform: translateY(2px);
   min-width: min-content;
-  width: sage-container(xs);
+  width: sage-container(modal);
   margin: 20vh 0 0;
   border-radius: sage-border(radius-large);
```
philschanely: I think the design is asking for `radius` (`4px`) now.

lib/sage-frontend/stylesheets/system/patterns/objects/_modal.scss
```diff
@@ -33,38 +33,63 @@ $-modal-padding-y: sage-spacing(md);
...
   background-color: sage-color(white);
   box-shadow: sage-shadow(lg);
```
philschanely: Looks like the design is using `2xl` shadow now.

lib/sage-frontend/stylesheets/system/patterns/objects/_tabs.scss
```diff
@@ -34,12 +36,82 @@
...
   }
 }
+
+.sage-tabs.sage-tabs--style-radio .sage-tabs__tab {
+  @include button-style-reset();
+  display: flex;
+  flex-grow: 1;
+  flex-basis: 0;
+  align-items: center;
+  height: rem(60);
```
philschanely: According to latest type spec, the line-height of the element should be 28px with top/bottom padding of 16px. Would you be open to switching to this padding rather than fixing the height? This plus your consideration of line clamping may be more flexible moving forward.

lib/sage-frontend/stylesheets/system/patterns/objects/_tabs.scss
```diff
@@ -34,12 +36,82 @@
...
+    transition: border $sage-radio-transition;
+    display: inline-block;
+    border-radius: sage-border(radius-round);
+    border: sage-border();
+    margin: 0 sage-spacing(xs) 0 0;
+    box-sizing: border-box;
+  }
+
+  &:hover {
+    background-color: sage-color(grey, 200);
```
philschanely: I don't see this BG color in the design spec; are they on board with this interpretation over the spec?

Screen Shot 2020-07-21 at 11 30 21 AM

lib/sage-frontend/stylesheets/system/patterns/objects/_tabs.scss
```diff
@@ -34,12 +36,82 @@
...
+  border-radius: sage-border(radius);
+  margin: 0 sage-spacing(sm) 0 0;
+  transition: color $sage-radio-transition, background-color $sage-radio-transition, box-shadow $sage-radio-transition;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
+
+  @extend %t-sage-body-med;
+
+  &:before {
```
philschanely: I'm getting some responsive squishing on these radios. :-/

Screen Shot 2020-07-21 at 11 31 10 AM

lib/sage-frontend/stylesheets/system/patterns/objects/_tabs.scss
```diff
@@ -34,12 +36,82 @@
...
 }
+
+.sage-tabs.sage-tabs--style-radio .sage-tabs__tab {
+  @include button-style-reset();
+  display: flex;
+  flex-grow: 1;
+  flex-basis: 0;
+  align-items: center;
+  height: rem(60);
+  color: $sage-tabs-default-color;
```
philschanely: I'm seeing `charcoal, 400` for all states of this text color. Might be worth double-checking with Design how much they want tab styles aligned here.

lib/sage-frontend/stylesheets/system/patterns/objects/_tabs.scss
```diff
@@ -6,11 +6,13 @@
 
 .sage-tabs {
   display: flex;
-  margin-bottom: sage-spacing();
+  margin-bottom: $sage-inputfield-spacing;
 }
 
-.sage-tabs__tab {
+.sage-tabs:not(.sage-tabs--style-radio) .sage-tabs__tab {
```
philschanely: Perhaps we can add a selector for `.sage-tabs__tab` and nest this as well as line 48 under it. It also seems that `.sage-tabs` is not needed before the `:not()` or the `.sage-tabs--style-radio`.

Thanks for this thought to guard the styles and thus avoid too many overrides! 🙌

lib/sage-frontend/stylesheets/system/patterns/objects/_tabs.scss
```diff
@@ -34,12 +36,82 @@
   }
 }
 
-.sage-tabs__tab--active {
+.sage-tabs:not(.sage-tabs--style-radio) .sage-tabs__tab--active {
```
philschanely: Similar to above, this could be nested inside `.sage-tabs__tab--active` along with line 100 and the `.sage-tabs` is not needed at the front.

app/views/examples/objects/tabs/_markup.html.erb
```diff
@@ -1,5 +1,23 @@
...
-  <a href="#" class="sage-tabs__tab <%= option[:is_active] ? "sage-tabs__tab--active" : "" %>"><%= option[:text] %></a>
+    <a href="#" data-js-tabs-target="<%= option[:id] %>" class="sage-tabs__tab"><%= option[:text] %></a>
   <% end %>
-</div>
\ No newline at end of file
+</div>
+
+<div class="sage-tabs" data-js-tabs>
+  <% options.each do |option| %>
+    <button type="button" data-js-tabs-target="<%= option[:id] %>" class="sage-tabs__tab"><%= option[:text] %></button>
```
philschanely: Focus styles differ between hyperlink and buttons here. Might be worth tidying up.

app/views/examples/objects/panel/_markup.html.erb
```diff
@@ -1,5 +1,10 @@
-<div class="sage-panel">
-  <div class="sage-panel__header"><%= header %></div>
+<section class="sage-panel">
```
philschanely, line 1: Is there a design spec for this object?

philschanely (changes requested):
Fabulous work on the JS! I looked with 🔍  at the new/updated CSS and markup so sorry in advance for all the nit-picking. 🤓 🙇‍♂️  

Excited about this work for sure!! 🙌

lib/sage-frontend/stylesheets/system/patterns/objects/_tabs.scss
AndrwM: Great idea! Thank you

lib/sage-frontend/stylesheets/system/patterns/objects/_tabs.scss
AndrwM: Ooops, fixing!

lib/sage-frontend/stylesheets/system/patterns/objects/_tabs.scss
AndrwM: I'll double check on this one

app/views/examples/objects/panel/_markup.html.erb
AndrwM: I've been working off this for the spec: https://www.figma.com/file/XqyDa4KJg8qKy62HULoNyA/Navigation?node-id=169%3A142

This is a screengrab of Kajabi-Products:
image

AndrwM: This spacing is a little unconventional to take into account the type of content that goes into the containers.
- footer has less x-padding to allow for the button text to align with the title
- header-aside does the same but on the y-axis

philschanely: I see your point. I wonder though whether Design considered the hover and focus effects and how they would impact this spacing. 

It might be worth some fancier styling to set up cleaner padding by default but then to perhaps just offset the particular buttons to space as desired? This way perhaps the component is more content-ready for different possibilities while also accounting for this particular implementation?

AndrwM: I love that idea. I'll experiment with a button helper to account for this.

I brought this up with @croswell a few weeks back and let him know that the spacing was irregular but that I would see what I could do.

Since this isn't an official pattern in the spec it might be worth a chat with several of the Sage Design stakeholders to ask for more example comps of Admin using the Panel so that we can figure out a standard approach to padding.

lib/sage-frontend/stylesheets/system/patterns/objects/_tabs.scss
AndrwM: I went ahead and reduced it, it was a bit heavy

app/views/examples/objects/modal/_markup.html.erb
```diff
@@ -1,14 +1,14 @@
-<div class="sage-modal" data-js-modal="cool-modal" role="dialog" aria-modal="true">
-  <div class="sage-modal__container">
-    <div class="sage-modal__header">
-      <h5 class="t-sage-heading-5">Example Sage Modal</h5>
-    </div>
+<dialog class="sage-modal" data-js-modal="cool-modal" role="dialog" aria-modal="true">
```
teenwolfblitzer: Interesting! I had thought `dialog` support was still not fully rolled out in Safari, but it appears to be working just fine. I would assume that we'd no longer need `role="dialog"` though, since the element itself has that role.

app/views/examples/objects/modal/_markup.html.erb
```diff
@@ -1,14 +1,14 @@
...
+  <section class="sage-modal__container">
+    <header class="sage-modal__header">
+      <h1 class="t-sage-heading-4">Example Sage Modal</h1>
+    </header>
     <div class="sage-modal__content">
       <p class="t-sage-body">Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.</p>
     </div>
     <div class="sage-modal__footer">
+      <a href="#" class="sage-btn sage-btn--tertiary" data-js-modal-close>Close Modal</a>
```
teenwolfblitzer: Hmm… I'm thinking both of these could be `button` elements just for the sake of this example.

app/views/examples/objects/tabs/_markup.html.erb
```diff
@@ -1,5 +1,23 @@
...
+  <% end %>
+</div>
+
+<div class="sage-tabs sage-tabs--style-radio" data-js-tabs>
```
teenwolfblitzer: This is looking great! As a follow-up (not a blocker for this PR), I think we'll want to add in `id`s and `aria-controls` to tie the tabs + panels together, similar to [the recommended implementation](https://www.w3.org/TR/wai-aria-practices/examples/tabs/tabs-1/tabs.html)

lib/sage-frontend/javascript/system/init.js
```diff
@@ -1,79 +1,95 @@
...
+        unbindFunc(this);
+      });
+    }
+  }
+
+  let initDocumentEventListener = function(eventName, handlerFunc) {
+    document.addEventListener(eventName, handlerFunc);
+  }
+
+  initDocumentPresenceListener('[data-js-modal]',           Sage.modal.init,            false);
```
teenwolfblitzer, line 36: 😎

lib/sage-frontend/javascript/system/modal.js
```diff
@@ -3,42 +3,64 @@ Sage.modal = (function() {
...
+  const EVENT_CLOSEALL = 'sage.modal.closeAll';
 
   // ==================================================
   // Functions
   // ==================================================
-  function init() {
-    modalTriggers.forEach(function(el) {
-      el.addEventListener('click', handleTriggerClick);
+
+  function init(el) {
```
teenwolfblitzer, line 16: Question for the group @philschanely @pixelflips , since this reminds me of a comment that Phil brought up on one of my PRs (and also because of Andrew's [comment below](https://github.com/Kajabi/sage/pull/331/files#r458160291)). I like that @AndrwM has already started the convention of setting module vars in all caps for example, and wonder if we can move forward with unifying naming conventions on JS variables functions, and arguments. `el` or `ele`/`elem` and `e` vs `evt` or `event`, etc. Standardizing our naming conventions here seems like a natural step since we're doing the same with CSS and BEM.

lib/sage-frontend/stylesheets/system/patterns/objects/_dropdown.scss
```diff
@@ -120,7 +127,7 @@ $-dropdown-item-focus-line-spacing-x: sage-spacing(sm);
   &:active {
```
teenwolfblitzer: Can this be combined with the `focus` styles above? the `border-bottom-color` could then be split out as a separate piece if necessary.

lib/sage-frontend/stylesheets/system/patterns/objects/_panel.scss
```diff
@@ -16,11 +16,11 @@
...
-  padding: sage-spacing(xs) sage-spacing(xs) 0 $sage-panel-padding;
+  padding: $sage-panel-padding;
   border-bottom: sage-border();
 }
 
-.sage-panel__header-right {
+.sage-panel__header-aside {
   display: inline-flex;
   align-self: baseline;
   margin: 0 0 0 auto;
```
teenwolfblitzer: I think we can reduce this one down to just `margin-left: auto` so we won't have to override later

lib/sage-frontend/stylesheets/system/patterns/objects/_tabs.scss
```diff
@@ -6,40 +6,115 @@
...
+  white-space: nowrap;
   color: $sage-tabs-default-color;
-  transition: color $sage-tabs-default-transition;
+  transition: color $sage-radio-transition, background-color $sage-radio-transition, box-shadow $sage-radio-transition;
 
   @extend %t-sage-body-med;
 
-  &:last-of-type {
-    margin-right: 0;
+  :not(.sage-tabs--style-radio) > & {
```
teenwolfblitzer, line 26: I suspect we may want to refactor this block in the near future since it feels a little too difficult to parse to me and the nesting gets pretty deep 🔦

teenwolfblitzer (approved):
👍  🔥  LGTM, some really great work! A few items to look through and a question for the group; nothing I can see that would block a merge.

lib/sage-frontend/javascript/system/modal.js
philschanely: Perhaps add a list of our conventional names somewhere in our READEM or docs?

lib/sage-frontend/stylesheets/system/patterns/objects/_tabs.scss
philschanely: I see your point @monkeypox8; @AndrwM already reduced the complexity here a bit and I do appreciate the :not() that guards some of the default styling from having to be overridden later. On the other hand its visually different enough it may warrant a different object altogether. What do you think?

philschanely (approved):
Thanks for working through all the critique. 🙇‍♂️  🏆

lib/sage-frontend/stylesheets/system/patterns/objects/_tabs.scss
AndrwM: @monkeypox8 @philschanely –
I've been struggling with whether to split it off into its own object. The idea that I keep coming back to is that the functionality and implementation in Products is the same.

They both:
- Use the same JS
- The same `.sage-tabs-pane` class
- Are implemented in Products as:
   ```
    <%= sage_component SageTabs, { style: "radio", items: ... } %>

    <%= sage_component SageTabsPane, { id: ... } do %> (block) <% end %>
   ```

I agree that the tabs scss is pretty hard to parse, I wonder if there is a clever abstraction we can use here. (Individual scss files for each Tab style? – I'm not sure).

lib/sage-frontend/javascript/system/modal.js
AndrwM: I'll add this in a followup and ping you all for review. 👍

lib/sage-frontend/stylesheets/system/patterns/objects/_dropdown.scss
AndrwM: Good idea 👍 Updated

lib/sage-frontend/stylesheets/system/patterns/objects/_tabs.scss
AndrwM: I refactored this to use pseudo selector variable names for better readability.

PR #329, opened 2020-07-14
Sidenav style updates

teenwolfblitzer:
## Description

- Revises sidebar nav/link styles to match [updated Figma comps](https://www.figma.com/file/xlUFoRTjsFePavVT85SmLL/Sage-Design-System-v1.0?node-id=2322%3A6093)
- Sidebar and content backgrounds updated to more closely align with `kajabi-products` styling

|  In use   |  Display states  |
|--------|--------|
||sidenav-states|

|  before   |  after  |  app view |
|--------|--------|--------|
|sidenav-before|sidenav-after|app-current|

## Related

- n/a

### DIFF
--- a/lib/sage-frontend/stylesheets/system/layout/_content.scss
+++ b/lib/sage-frontend/stylesheets/system/layout/_content.scss
@@ -8,6 +8,6 @@
   flex: 1;
   overflow: auto;
   padding: sage-spacing() 0;
-  background: sage-color(grey, 100);
+  background: sage-color(grey, 200);
   scroll-behavior: smooth;
 }
--- a/lib/sage-frontend/stylesheets/system/patterns/objects/_nav.scss
+++ b/lib/sage-frontend/stylesheets/system/patterns/objects/_nav.scss
@@ -3,6 +3,9 @@
   type: object
 
 ================================================== */
+$-sage-nav-color-text: sage-color(charcoal, 400);
+$-sage-nav-color-focus: sage-color(primary);
+$-sage-nav-color-background: sage-color(primary, 400);
 
 .sage-nav__list {
   margin: 0;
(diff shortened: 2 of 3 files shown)


pixelflips (approved):
🚀   - LGTM!

AndrwM (approved):
Nice! This looks great.

I wonder what the team thinks about having some type of "Pro"/"Compact" spacing/text-sizing for this that can be used within Super Admin and Sage Docs. The amount of links that we display on Admin compared to the others is significantly less. It's starting to feel like the Sage Docs are a bit unusable with all the links we're stuffing in there.

philschanely (approved):
Nice! 🙌

philschanely:
> Nice! This looks great.
> 
> I wonder what the team thinks about having some type of "Pro"/"Compact" spacing/text-sizing for this that can be used within Super Admin and Sage Docs. The amount of links that we display on Admin compared to the others is significantly less. It's starting to feel like the Sage Docs are a bit unusable with all the links we're stuffing in there.

Perhaps we don't need to display the submenus under Elements and Objects anymore. They also both have the on-page navigation on the right side?

teenwolfblitzer:
> > Nice! This looks great.
> > I wonder what the team thinks about having some type of "Pro"/"Compact" spacing/text-sizing for this that can be used within Super Admin and Sage Docs. The amount of links that we display on Admin compared to the others is significantly less. It's starting to feel like the Sage Docs are a bit unusable with all the links we're stuffing in there.
> 
> Perhaps we don't need to display the submenus under Elements and Objects anymore. They also both have the on-page navigation on the right side?

👍 I like the idea of removing the submenus from the overviews of both, since the list is bound to grow even longer in the near future. I'll open a follow-up issue to investigate.

PR #337, opened 2020-07-15
Hotfix: Update Sage Form Section padding

AndrwM:
## Description

This is a regression from the Sage Panel markup update. The Sage Form Section depended on an extend of `%sage-panel`, since Sage Panel's padding has moved to its child element, `.sage-panel__body`, the Sage Form Section was left padding-less.

**This only affects Sage Super Admin**

### Screenshots

|  before  |  after  |
|--------|--------|
|image|image|


pixelflips (approved):
👍  - LGTM!

PR #338, opened 2020-07-17
Sage Docs: Rails Routing & Navigation Bars

### LINKED ISSUE #336: Investigate removal of submenu items in sidebar
## Description

We have a couple of related issues happening with the sidebar, most likely due to the logic determining where/when to display subitems within a parent group.

1. The submenu lists of individual elements and objects are much longer than the intended use within the admin, and are also duplicated in the "Quick links" menu.
2. "Styles" subitems do not render when selecting an item from the menu. For example, selecting "Colors" no longer indicates the page is part of "Styles". This is likely due to how pages in Styles do not follow the same URL pattern as "Elements" and "Objects".

Action items:
- [ ] Investigate "Styles" section path and/or sidebar logic
- [ ] Explore impact of removing the submenu links from the sidebar
- OPTIONAL: look into using Intersection Observer or some other scroll detection method to highlight current position within the Quick links menu

## Related

- #329

AndrwM:
## Changelog
- Update routing to use paths instead of query string params to look up elements/objects/breakouts
- Refactor how we're checking if a page or child page is 'active'
  - Clean up some conventions (like `current?` helper) and take a more "Rails" approach with built in helpers
  - Fixes the collapsing sidebar bug
- Docs html tidying up

Paths now follow this convention:
```
/pages/style/token
/pages/elements
/pages/element/button
/pages/objects
/pages/object/button
/pages/breakout/object/button

```

Closes: #336


app/views/application/_sidebar.html.erb
```diff
@@ -2,27 +2,28 @@
   <div class="sage-sidebar__body">
     <nav class="sage-nav" aria-label="Main">
       <ul class="sage-nav__list">
-        <li><%= link_to "Introduction", pages_index_url, class: "#{current? "sage-nav__link--active", pages_index_url} sage-nav__link" %></li>
-        <li><%= link_to "Styles", pages_token_url, class: "sage-nav__link" %>
-          <% if current_page?(pages_token_url) %>
-          <ul class="sage-nav__list sage-nav__list--sub">
-            <%= yield :sidebar_styles %>
-          </ul>
+        <li><%= link_to "Introduction", pages_index_path, class: "sage-nav__link #{"sage-nav__link--active" if current_page?(pages_index_path)}" %></li>
```
teenwolfblitzer, line 5: One thing I was hoping to get some discussion around is our use of classes on the link, but not on its parent `li`. I could see a case for `sage-nav__list-item` and adding the `active` style there. That's not to say it's a result of this work or that any potential change should be, either.

lib/sage-frontend/stylesheets/docs/_sidebar.scss
```diff
@@ -17,3 +17,9 @@
     opacity: 1;
   }
 }
+
+// Side Navigation Footer Top Border
+// Visual division between the scrolling content and the footer
+.sage-sidebar__footer {
+  box-shadow: inset 0px 1px 0px sage-color(gray);
```
teenwolfblitzer: I'm being insanely nitpicky here, but we can drop the `px` units for zero :neckbeard:

teenwolfblitzer (approved):
👍 🔥 Looks great! Couple of discussion points, but nothing that would hold this up. Great work!

lib/sage-frontend/stylesheets/docs/_sidebar.scss
AndrwM: Haha, of course!

app/views/application/_sidebar.html.erb
AndrwM: Yeah, good point. The `<li>` should have some type of class. 👍

PR #340, opened 2020-07-20
New icon font system

philschanely:
## Description
This PR updates the icon font to use a new setup exported from IcoMoon that seems to be a much simpler process for Design to maintain and add new icons to moving forward. 

This also incorporates a new sizing system based on a 4pt incremental grid. `md` is the default at 16px width and height, so `lg` is 20px, `xl` is 24px, `2xl` is 28px, and so forth. See `$sage-icon-sizes` in the Icons token file for more.

A handful of new icons are also added with this PR.


lib/sage-frontend/stylesheets/system/core/_mixins.scss
```diff
@@ -152,12 +152,24 @@
...
+  width: $size;
+  height: $size;
   text-transform: none;
-  font: normal normal normal 1em/1 "Sage";
+  font-family: "Sage";
+  font-size: $size;
+  font-style: normal;
+  font-weight: normal;
+  line-height: $size;
   speak: none;
```
teenwolfblitzer: Should this be `speak: never`? I'm not 100% sure when the spec changed, but all recent references appear to be pointing that direction.

lib/sage-frontend/stylesheets/system/core/_icons.scss
```diff
@@ -15,12 +15,12 @@ $sage-icon-li-margin-right: 0.4em !default;
   font-family: "Sage";
   src: url("#{$sage-icon-font-path}/Sage.eot");
   src: url("#{$sage-icon-font-path}/Sage.eot") format("embedded-opentype"),
```
teenwolfblitzer: @pixelflips may know more about this from his recent web perf workshop, but it's my understanding that source order can make some difference here: should we be prioritizing `woff` since it's the 'newest' spec/most well-supported?

I _think_ we can also drop `eot` since it's primarily intended for < IE9, but it wouldn't be likely to load on other browsers if we kept it in, lower in the src list.

teenwolfblitzer (approved):
👍 💥  LGTM, a couple of non-blocking items to consider. Nice work!

lib/sage-frontend/stylesheets/system/core/_icons.scss
philschanely: Sweet. Good point.

philschanely: @pixelflips ☝️

pixelflips: @monkeypox8 - nice catch and src order can make a difference, but mostly around the actual download size if I remember correctly. Some fonts types are compressed better than others.

@philschanely I'm good with the ordering you suggested, but any reason we lost woff2, just curious?

philschanely: Sadly the icomoon system did not provide one :-(

pixelflips: > Sadly the icomoon system did not provide one :-(

@philshanely - got it. thanks for the info and nice work!

PR #349, opened 2020-07-23
Copy button

philschanely:
## Description
Adds a `data-js-copy-button` attr that can be added to a button in order to copy text to the clipboard (see Sandbox page now; will remove before merging).

### DIFF
--- a/app/views/pages/sandbox.html.erb
+++ b/app/views/pages/sandbox.html.erb
@@ -3,4 +3,4 @@
 <% end %>
 <p>Use this scratchpad to test out elements and objects in context. Just be sure not to commit your changes. ❤️</p>
 
-<button class="sage-btn sage-btn--primary sage-btn--icon-right-copy" data-js-copy-button="abra cadabra">Copy</button>
+<button class="sage-btn sage-btn--primary sage-btn--icon-right-copy" data-js-copy-button="abra cadabra">Copy secret</button>
--- a/app/views/pages/sandbox.html.erb
+++ b/app/views/pages/sandbox.html.erb
@@ -2,3 +2,5 @@
   <h1>Sandbox</h1>
 <% end %>
 <p>Use this scratchpad to test out elements and objects in context. Just be sure not to commit your changes. ❤️</p>
+
+<button class="sage-btn sage-btn--primary sage-btn--icon-right-copy" data-js-copy-button="abra cadabra">Copy</button>
--- a/app/views/pages/sandbox.html.erb
+++ b/app/views/pages/sandbox.html.erb
@@ -4,3 +4,7 @@
 <p>Use this scratchpad to test out elements and objects in context. Just be sure not to commit your changes. ❤️</p>
 
 <button class="sage-btn sage-btn--primary sage-btn--icon-right-copy" data-js-copy-button="abra cadabra">Copy secret</button>
+
+<button class="sage-btn sage-btn--tiny sage-btn--tertiary sage-btn--icon-right-copy" data-js-copy-button="abra cadabra" data-js-tooltip="Copy">
+  <span class="sage-label sage-label--draft">abra cadabra</span>
+</button>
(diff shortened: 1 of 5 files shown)


app/views/pages/sandbox.html.erb
```diff
@@ -1,4 +1,10 @@
...
 <% end %>
-<p>Use this scratchpad to test out elements and objects in context. Just be sure not to commit your changes. ❤️</p>
\ No newline at end of file
+<p>Use this scratchpad to test out elements and objects in context. Just be sure not to commit your changes. ❤️</p>
+
+<button class="sage-btn sage-btn--primary sage-btn--icon-right-copy" data-js-copy-button="abra cadabra">Copy secret</button>
+
+<button class="sage-btn sage-btn--tiny sage-btn--tertiary sage-btn--icon-right-copy" data-js-copy-button="abra cadabra" data-js-tooltip="Copy">
+  <span class="sage-label sage-label--draft">abra cadabra</span>
+</button>
```
AndrwM: Don't forget to remove :)

lib/sage-frontend/javascript/system/copy-button.js
```diff
@@ -0,0 +1,40 @@
...
+
+  function init() {
+    const buttons = document.querySelectorAll(COPY_BUTTON_ATTR);
+
+    buttons.forEach(button => {
+      button.addEventListener('click', (ev) => {
+        copyToClipboard(ev.currentTarget.dataset.jsCopyButton);
+      });
+    });
+  }
```
AndrwM: If you're using the "Presence" init strategy, an argument is passed to the `init` that is a reference to the element in the DOM. In this current state, if there were 4 instances on the page this will be run 4 times (and therefore be bound 16 times 😉 ). 

When you refactor this would you mind unbinding too? I'd love for us to get into the habit of doing it (even though the performance overhead for this is in particular is pretty minute).

See `Sage.dropdown.init` and `Sage.dropdown.unbind` as examples.

AndrwM (changes requested):
This looks great! Made some comments on the JS init. Lemme know if you wanna chat IRL about it!

teenwolfblitzer (approved):
👍  Works great! Seems like it'd be pretty easy to also extend this functionality as a util for textareas, etc down the road as well 😎

lib/sage-frontend/javascript/system/copy-button.js
teenwolfblitzer: @AndrwM Is the mutationsobserver listener only for unbinding event listeners on the document?

AndrwM: @monkeypox8 it's also for binding.

When a DOM element arrives that has the correct data-attr then it will be passed to `init` to be set up. Then when it is removed from the DOM the instance will be passed to `unbind`.

AndrwM: @monkeypox8 Here are the docs: https://github.com/uzairfarooq/arrive

😄

teenwolfblitzer: Whoops, my wording there was unclear. I'd intended to ask whether arrive's removing an element would also require manually unbinding its event listeners but that seems to be covered by the linked docs (and is likely good general practice anyway)

app/views/pages/sandbox.html.erb
philschanely: Will do; left it in for testers' benefit. Pulled now.

philschanely:
> This looks great! Made some comments on the JS init. Lemme know if you wanna chat IRL about it!

Thanks, @AndrwM ! Does this look as you'd expect now?

lib/sage-frontend/javascript/system/copy-button.js
```diff
@@ -0,0 +1,38 @@
+Sage.copyButton = (function() {
+
+  // ==================================================
+  // Variables
+  // ==================================================
+
+  function handleClick(ev) {
```
AndrwM, line 7: Can we move this into the "Functions" section?

AndrwM (approved):
Awesome, thanks for making these changes! +1

PR #350, opened 2020-07-23
Accordion panel

philschanely:
## Description

Modifies panel to allow an accordion variation. This is MVP with room to ❤️  on it for some improved transition polish.

### Screenshots

Screen Shot 2020-07-23 at 1 04 24 PM

## Test notes

Check out the new panel under Objects.

### DIFF
--- a/lib/sage-frontend/javascript/system/accordion-panel.js
+++ b/lib/sage-frontend/javascript/system/accordion-panel.js
@@ -1,0 +1,100 @@
+Sage.accordion = (function () {
+  // ==================================================
+  // Variables
+  // ==================================================
+
+  const JS_ACCORDION_ROOT = 'data-js-accordion';
+  const JS_ACCORDION_BODY = 'body';
+  const JS_ACCORDION_HEADER = 'header';
+  const JS_ACCORDION_SOLO_TOGGLE = 'group-toggle-solo';
+  const JS_ACCORDION_ID_BODY_PREFIX = 'accordion-body';
+  const JS_ACCORDION_ID_HEADER_PREFIX = 'accordion-header';
+
+  const relevantKeys = {
+    enter: 13,
+    space: 32,
+  };
+
+
+  // ==================================================
+  // Functions
+  // ==================================================
+
+  function generateId(prefix) {
+    const hash = Math.random().toString(36).substring(2, 15) + Math.random().toString(36).substring(2, 15);
+    return `${prefix}-${hash}`;
+  }
+
+  function handleAccordionClick(ev) {
+    const el = ev.target;
+
+    // Ensure relevant events before proceding
+    const isClick = ev.type === 'click';
+    const isRelevantKeypress = ev.type === 'keypress' && Object.values(relevantKeys).indexOf(ev.which) >= 0;
+    if (!isClick && !isRelevantKeypress) {
+      return false;
+    }
+
+    // Check for other items in an exclusive panel group to toggler off
+    const grandparent = el.parentNode.parentNode;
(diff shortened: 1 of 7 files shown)


lib/sage-frontend/javascript/system/accordion-panel.js
```diff
@@ -0,0 +1,97 @@
...
+    const hash = Math.random().toString(36).substring(2, 15) + Math.random().toString(36).substring(2, 15);
+    return `${prefix}-${hash}`;
+  }
+
+  function handleAccordionClick(ev) {
+    const el = ev.currentTarget;
+
+    // Ensure relevant events before proceding
+    const isClick = ev.type === 'click';
+    const isRelevantKeypress = ev.type === 'keypress' && Object.values(relevantKeys).indexOf(ev.which) >= 0;
```
teenwolfblitzer, line 33: 🆒 Very nice, I could see this and the keymaps obj above eventually becoming a util function

lib/sage-frontend/stylesheets/system/patterns/objects/_panel.scss
```diff
@@ -33,6 +64,28 @@
...
+    &::after,
+    &::before {
+      content: "";
+      display: block;
+      height: $sage-panel-padding;
+    }
+  }
+
+  [aria-expanded="false"] + & {
+    height: 0;
```
teenwolfblitzer, line 84: Follow-up item: rather than animating height, I think we should consider using `transform: scale` on this to squeeze out a bit more performance from the animation. That'd probably mean an additional wrapper around the content if `p` tags are optional, however.

lib/sage-frontend/stylesheets/system/patterns/objects/_panel.scss
```diff
@@ -4,20 +4,51 @@
...
-  min-height: rem(60px);
   padding: $sage-panel-padding;
   border-bottom: sage-border();
+
+  .sage-panel--accordion & {
+    padding-top: sage-spacing(xs);
+    padding-bottom: sage-spacing(xs);
+    cursor: pointer;
+
+    &::after {
```
teenwolfblitzer, line 32: Stubborn ol' Voiceover insists on reading out the pseudo-element arrow icon, and I don't know that there's a way around it in CSS (yet). I suppose we could set the icon in a `span` with `aria-hidden="true"` if it's a concern. I think for now the readability of the button text alone would enough to pass muster.

lib/sage-frontend/stylesheets/system/patterns/objects/_panel.scss
```diff
@@ -4,20 +4,51 @@
...
+      @include icon-base(right-small, xl);
+
+      transform: translateX(rem(3px));
+      transform-origin: center;
+      margin: 0 0 0 auto;
+      color: sage-color(charcoal, 100);
+      transition: $-panel-transition transform;
+    }
+
+    &:focus {
```
teenwolfblitzer, line 43: Thoughts on including an `:active` style here as well?

teenwolfblitzer (approved):
👍 💎 LGTM, nice work!

lib/sage-frontend/stylesheets/system/patterns/objects/_panel.scss
philschanely: Got it!

lib/sage-frontend/stylesheets/system/patterns/objects/_panel.scss
philschanely: Good point. Great for a follow up indeed for further polishing

PR #354, opened 2020-07-24
Patch accordion panel omission

philschanely:
## Description
Patch accidental omission of accordion panel initialization


pixelflips (approved):
👍  LGTM!

PR #355, opened 2020-07-27
Panelize Sage Docs

AndrwM:
- Adds Sage Panel `lg` modifier per direction from @croswell that we'll want it later on
- Converts all Elements and Objects pages to use Sage Panel
- Converts most other pages to use Sage Panel
- DRY up some partials

### DIFF
--- a/app/views/examples/elements/_element_preview.html.erb
+++ b/app/views/examples/elements/_element_preview.html.erb
@@ -5,7 +5,6 @@
     <h2 id="<%= permalink %>" class="t-sage-heading-3">
       <%= link_to title.titleize, "##{permalink}", class: "example__link" %>
     </h2>
-
     <div class="sage-panel__header-aside">
       <%= link_to pages_element_path(title), class: "sage-btn sage-btn--tertiary sage-btn--icon-right-preview-on", data: { "js-tooltip": "Inspect element details" } do %>
         Details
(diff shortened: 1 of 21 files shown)


app/views/examples/elements/_element_preview.html.erb
```diff
@@ -1,13 +1,20 @@
...
+      <%= link_to title.titleize, "##{permalink}", class: "example__link" %>
     </h2>
-    <%= link_to pages_breakout_path(type: :element, title: title), target: "_blank", class: "example__link example__link--breakout" do %>
-      <i class="sage-icon-launch" aria-hidden="true"></i><span class="visually-hidden">Opens <%= title.titleize %> in a new window</span>
-    <% end %>
+    <div class="sage-panel__header-aside">
+      <%= link_to pages_element_path(title), class: "sage-btn sage-btn--tertiary sage-btn--icon-right-preview-on", data: { "js-tooltip": "Inspect element details" } do %>
+        Details
+      <% end %>
+      <%= link_to pages_breakout_path(type: :element, title: title), target: "_blank", class: "sage-btn sage-btn--tertiary sage-btn--icon-only-launch", data: { "js-tooltip": "Open breakout page" } do %>
```
teenwolfblitzer, line 12: Curious what the group thinks about dropping the breakout page links, since the 'details' link serves the same purpose.

teenwolfblitzer (approved):
👍 🔥 Looking great, nice work!

app/views/examples/elements/_element_preview.html.erb
AndrwM: I like how clean the presentation is in the breakout links but at this point they aren't serving a lot of purpose. I dropped it on the agenda 👍

PR #356, opened 2020-07-28
Adjust link element external icon spacing

pixelflips:
## Description
When the link element is an external link the icon spacing needs adjustment on the right side for consistent spacing when the link is inline with other text. Currently the spacing is a bit tight on the right.

### Screenshots
Note: screenshots are from project currently in progress in kajabi-products

|  Before   |  After  |
|--------|--------|
| Screen Shot 2020-07-27 at 4 28 13 PM|Screen Shot 2020-07-27 at 4 58 52 PM| 

## Related issues
- n/a

### DIFF
--- a/lib/sage-frontend/stylesheets/system/patterns/elements/_link.scss
+++ b/lib/sage-frontend/stylesheets/system/patterns/elements/_link.scss
@@ -1,7 +1,8 @@
 .sage-link--launch,
 .sage-type a[target="_blank"]:not(.sage-link--no-launch) {
   &::after {
-    margin: 0 sage-spacing(2xs);
+    margin-left: sage-spacing(2xs);
+    margin-right: sage-spacing(2xs);
     @include icon-base(launch, sm);
   }
 }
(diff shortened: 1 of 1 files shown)


lib/sage-frontend/stylesheets/system/patterns/elements/_link.scss
```diff
@@ -1,7 +1,7 @@
 .sage-link--launch,
 .sage-type a[target="_blank"]:not(.sage-link--no-launch) {
   &::after {
-    margin-left: sage-spacing(xs);
+    margin: 0 sage-spacing(2xs);
```
teenwolfblitzer: Rather than the shorthand, which explicitly sets a zero-margin to top and bottom, could we use `margin-left` and `margin-right` here?

teenwolfblitzer (approved):
👍 LGTM, much nicer spacing! Small change to consider in order to prevent setting properties where we weren't previously, not a blocker.

lib/sage-frontend/stylesheets/system/patterns/elements/_link.scss
pixelflips: > Rather than the shorthand, which explicitly sets a zero-margin to top and bottom, could we use `margin-left` and `margin-right` here?

@monkeypox8 will adjust. thanks for pointing that out!

PR #357, opened 2020-07-28
Status Icon

AndrwM:
# Description
## New
- Sage Status Icon

## Polish
- Match Sage Tabs to the figma spec
- Fix Safari tooltip overflow
- Fix tooltip width issue with longer text
- Sage Sortable new drag handle icon
- Sage Dropdown `:active` underline color to match Sage Tabs

## Refactor
- Sage Card center, refactor how this is done to reduce paragraph line-length
- Add Sage Footer Aside for left aligned elements


lib/sage-frontend/stylesheets/system/patterns/elements/_tooltip.scss
```diff
@@ -8,7 +8,7 @@
   display: flex;
   position: absolute;
   z-index: $sage-tooltip-zindex;
-  width: max-content;
+  min-width: max-content;
```
teenwolfblitzer, line 11: 🥳

lib/sage-frontend/stylesheets/system/patterns/objects/_tabs.scss
```diff
@@ -19,7 +19,7 @@
   margin: 0 sage-spacing(sm) 0 0;
   white-space: nowrap;
   color: $sage-tabs-default-color;
-  transition: color $sage-radio-transition, background-color $sage-radio-transition, box-shadow $sage-radio-transition;
+  transition: color $sage-transition, background-color $sage-transition, box-shadow $sage-transition;
```
teenwolfblitzer, line 22: Small suggestion since these all share the same timing/easing, this can use `transition: $sage-transition` with `transition-property` on the individual properties. Makes changing timing & easing for (and adding new properties to transition) a little easier.

lib/sage-frontend/stylesheets/system/patterns/objects/_tabs.scss
```diff
@@ -30,16 +30,17 @@
     outline: none;
 
     &#{$item-underline} {
+      $spacing-off: 50%;
```
teenwolfblitzer, line 33: I'm wondering if this variable is necessary, since it's only for the two instances?

app/views/examples/elements/status_icon/_markup.html.erb
```diff
@@ -0,0 +1,3 @@
+<div class="sage-status-icon sage-status-icon--<%= icon_name %>" data-js-tooltip="<%= content %>">
```
teenwolfblitzer, line 1: We'll also want to include `aria-role="tooltip"` here to indicate the intent of the content.

As a follow-up item, any elements using the status icon should also be related with an `aria-describedby` pointing to the icon's `id`. Or maybe `aria-labelledby`. Depends on if we want it read after or before the content, respectively.

teenwolfblitzer (commented):
Nice work! Hidden text is working well as a fallback for voiceover on the tooltips. One change to look at on adding `aria-role`, the rest are non-blocking.

app/views/examples/elements/status_icon/_markup.html.erb
AndrwM: Good idea. I think `aria-role="tooltip"` should be owned by the js tooltip script.

I'll whip up a `buildA11y()` function within its `init()`.

lib/sage-frontend/stylesheets/system/patterns/objects/_tabs.scss
AndrwM: I'll remove 👍

lib/sage-frontend/stylesheets/system/patterns/objects/_tabs.scss
AndrwM: This makes sense to me. Updating

PR #359, opened 2020-07-30
Fix tabs click handler bug

AndrwM:
## Description
If the tabs parent is clicked it attempts to change the tab. Prevent those click events from taking an action.


pixelflips (approved):
LGTM +1 👍

PR #361, opened 2020-07-31
Sage System design review sizing adjustments

AndrwM:
## Description
A result of a Sage System design review with @mattpatty9 & @croswell in the context of the new Admin Navbars.

### System
- Modal: Header aside (placement for future top-right dropdown in modal)
- Spacing: Add a `2xl` spacing value of `64px`
- Checkbox: Shrink the checkboxes to `20px`
- Tabs: Use a larger margin for Tabs that aren't `radio` style
- Card: Add a large vertical padding modifier thats used for UI null state CTAs in Figma
- Panel: update bottom margin to `lg`


.gitignore
```diff
@@ -1,4 +1,5 @@
 .DS_Store
+.notes
```
AndrwM, line 2: Do you all mind if I add this to the gitignore? I've been keeping notes in a file like this recently.

pixelflips: > Do you all mind if I add this to the gitignore? I've been keeping notes in a file like this recently.

@AndrwM no problem here.

pixelflips (approved):
LGTM! 👍

app/views/examples/elements/_element_preview.html.erb
```diff
@@ -14,7 +14,7 @@
       <% end %>
     </div>
   </div>
-  <div class="sage-panel__body sage-type">
```
philschanely, line 17: Why are we removing this? It added default spacing to the contained type that helped the site present better IMO.

philschanely (approved):
One question about a change but non-blocking. 

Also be sure to run linter as well before merging ;)

app/views/examples/elements/_element_preview.html.erb
AndrwM: It's causing irregular spacing for a few nested typographic situations like this:
https://sage-design-system.kajabi.com/pages/object/sortable

I added this to our UXD agenda for us to discuss how we want to handle documentation type. I'm a fan of the `sage-type` pattern but I think it might need to be scoped down so it doesn't affect the entire preview.

philschanely: Agreed. You see the inverse is also a problem and I'll look forward to hashing out a better plan with the team. Thanks!

AndrwM: I'll go ahead and rip this change out for now, might be doing more harm than good for the moment.

philschanely: Its too bad either way that there's problems as a result. I do think the broader approach to documenting examples could be reconsidered. We essentially have many places where we're not just demonstrating one sample but also labeling variations. That supporting documentation may not have been a part of the original design pattern there. 🤷

AndrwM: Thats a great point. This is def a sign that we need to rethink how we're structuring variations in the docs.

PR #365, opened 2020-08-05
Add system-styles file

philschanely:
## Description

Adds a system-styles file to enable styles-only imports when needed by a dependent application.

Thanks @AndrwM


AndrwM (approved):
+1!

PR #366, opened 2020-08-05
Converts positioning of affix to fix value

philschanely:
## Description

(No visual change)

Due to additional elements that can be present or not inside a Sage Input element, the flexible positioning value of the prefix and suffixes needed to be converted to a reliable fixed value.


AndrwM (approved):
I'm okay with this. Are Sage Inputs intended to be variable height?

philschanely:
> I'm okay with this. Are Sage Inputs intended to be variable height?

Its more that when feedback or subtext appears it changes the height of the container and bumps the affix out of alignment with the field itself. :-/

PR #368, opened 2020-08-06
Add modal container token to docs and update spacing

AndrwM:
## Description
- Add modal container size token to docs
- Fix bothersome spacing when displaying long tokens

image


joe-dakroub (approved):
These changes look good to me.

AndrwM:
Let it rip – merging

PR #369, opened 2020-08-06
[MAN-365] Sage Tabs with style and stacked options

joe-dakroub:
## Description
We have need to display radio button options that can appear in a default (horizontal) or stacked (vertical) layout along with a corresponding pane that displays the content related to the chosen option.

The following changes were made to accommodate this new feature:
- Added a `choice` and `tab` Element that can be used independently
- Updated the `tabs` Object to use the new `tab` or `choice` Element depending on the configuration passed in
- Added a `stacked` prop to the tabs element
- Changed the `style` prop value to `choice` in place of `radio`
- Added `role="tablist"` to tabs container and `role="tabpanel"` to the tab panel
- Added `role="tab"` and `aria-controls` attribute to the tab and choice elements
- Added a toggle of the `aria-selected` attribute on the selected/deselected tab
- Added left/right cursor key support to navigate tabs

See https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role for more details

### Screenshots
SageRadioList

### Testing
1. You can test this locally: http://localhost:4000/pages/object/tabs

## Related
https://kajabi.atlassian.net/browse/MAN-365


lib/sage-frontend/stylesheets/system/patterns/objects/_tabs.scss
```diff
@@ -6,131 +6,25 @@
 
 .sage-tabs {
   display: flex;
-  margin-bottom: sage-spacing(lg);
+  margin-bottom: sage-spacing(sm);
 
-  &.sage-tabs--radio {
-    margin-bottom: $sage-inputfield-spacing;
+  &--layout-default {
```
philschanely: We prefer this to be a separate declaration rather than using `&` here.

lib/sage-frontend/stylesheets/system/patterns/objects/_tabs.scss
```diff
@@ -6,131 +6,25 @@
...
-        box-shadow: sage-shadow(sm);
-        border-width: 5px;
-        border-color: $sage-radio-color-checked;
-      }
-    }
-  }
-
-  &:last-of-type {
-    margin-right: 0;
+  &--layout-stacked {
```
philschanely: Same here.

philschanely (approved):
Small fix suggested below for convention but 🏆  otherwise! 🙇‍♂️

lib/sage-frontend/stylesheets/system/patterns/objects/_tabs.scss
joe-dakroub: Good catch! Resolved.

app/views/examples/objects/tabs/_markup.html.erb
```diff
@@ -1,21 +1,15 @@
...
-    </div>
-  <% end %>
+    <% end %>
+  </div>
 
   <% options.each do |option| %>
-    <div class="sage-tabs-pane" data-js-tabs-pane="<%= "#{id}-#{option[:id]}" %>">
-      Content for: <%= option[:text] %>
-    </div>
+    <div class="sage-tabs__pane" id="<%= "#{id}-#{option[:id]}" %>" data-js-tabs-pane="<%= "#{id}-#{option[:id]}" %>" role="tabpanel"><%= option[:content] || "" %></div>
```
AndrwM: With this naming schematic `sage-tabs__pane` should be a child node of `sage-tabs` to follow BEM syntax structure

app/views/examples/objects/tabs/_markup.html.erb
```diff
@@ -1,21 +1,15 @@
...
-    </div>
-  <% else %>
-    <div class="sage-tabs" data-js-tabs="<%= id %>">
-      <% options.each do |option| %>
-        <button type="button" data-js-tabs-target="<%= "#{id}-#{option[:id]}" %>" class="sage-tabs__tab"><%= option[:text] %></button>
-      <% end %>
-    </div>
-  <% end %>
+    <% end %>
+  </div>
```
AndrwM, line 10: I love the idea of having individual elements for a `tab` or a `choice` and extrapolating the interactivity into a parent object. This is NICE.

**2 Notes Here:**
- What does everyone think about renaming `sage-tabs` to something a bit more universal? On initial review it was confusing to me that this object to could accept `choice` and `tab` as children. Possibly:
```
sage-pane-controller
  sage-pane-controller__options
    sage-choice/tab
    sage-choice/tab
  sage-pane-controller__pane#name-of-pane1
  sage-pane-controller__pane#name-of-pane2
```
- Design has different bottom-margin requirement for "Tabs" (spacing `lg`) and "Choice" (spacing `md`). We should add some additional margin bottom to these `sage-tabs` (spacing `sm`, I think should suffice) to account for that additional spacing.

lib/sage-frontend/javascript/system/tabs.js
```diff
@@ -40,15 +43,42 @@ Sage.tabs = (function() {
...
+          selectedTarget = evt.target.nextElementSibling || evt.target.parentElement.firstElementChild
+        break
+      }
+
+      if (selectedTarget) {
+        dispatchChange(this.getAttribute(SELECTOR_TABS), selectedTarget.getAttribute(SELECTOR_TAB_ITEM));
+      }
+    }
+  }
+
```
AndrwM: 😍 👏

lib/sage-frontend/stylesheets/system/patterns/elements/_choice.scss
```diff
@@ -0,0 +1,101 @@
...
+
+  &:active,
+  &.sage-tab--active {
+    color: $sage-tabs-active-color;
+    &#{$item-radio} {
+      box-shadow: sage-shadow(sm);
+      border-width: 5px;
+      border-color: $sage-radio-color-checked;
+    }
+  }
```
AndrwM: `sage-tab--active` seems like a leak to me. What's our strategy for child elements being affected by parent objects? (cc @Kajabi/ux-devs )

Since this is all handled by a single JS script (currently named `sage-tabs`). Why don't we have that script detect whether the element that it is affecting is a `sage-tab` or `sage-choice` and apply the appropriate active class? ----> `sage-tab--active`/`sage-choice--active`

lib/sage-frontend/stylesheets/system/patterns/elements/_choice.scss
```diff
@@ -0,0 +1,101 @@
...
+
+    &:last-of-type {
+      margin-bottom: 0;
+    }
+
+    &#{$item-radio} {
+      margin-right: sage-spacing(sm);
+    }
+  }
+}
```
AndrwM: Same concern as above regarding BEM leaks. I'm not sure the best way to work around this without adding a modifier class to each choice.

AndrwM: How about using a CSS grid for the `sage-tabs` object?

The `grid-gap` property may give us what we need.

lib/sage-frontend/stylesheets/system/patterns/elements/_tab.scss
```diff
@@ -0,0 +1,66 @@
...
+      left: 0;
+      right: 0;
+    }
+  }
+
+  &:last-of-type {
+    margin-right: 0;
+  }
+}
+/* stylelint-enable max-nesting-depth */
```
AndrwM: I don't think we need to disable the nesting-depth lint here.

app/views/examples/objects/tabs/_props.html.erb
```diff
@@ -4,3 +4,15 @@
...
+  <td>stacked</td>
+  <td>Tabs use a <code>--stacked</code> modifier to visually display tab options on top of each other.</td>
+  <td>Boolean</td>
+  <td>false</td>
+</tr>
+<tr>
+  <td>style</td>
+  <td>Tabs use a <code>--style</code> modifier to visually display either a tab or choice button.</td>
+  <td>String</td>
+  <td>null</td>
```
AndrwM, line 17: This style modifier doesn't exist anymore. Please remove :)

AndrwM (changes requested):
Really awesome stuff. I like that we're separating the `tab` and `choice` into their own elements. However I'm concerned about a few BEM leaks happening and think that we could rename `sage-tabs` to something a bit more descriptive of its use.

lib/sage-frontend/javascript/system/tabs.js
joe-dakroub: It was a quick win since the JS code was so easy to work with!

app/views/examples/objects/tabs/_markup.html.erb
joe-dakroub: These are really good points @AndrwM. I struggled with this too and did have some thought about possibly renaming `sage-tabs` to something more generic. I could also see renaming `tab` and `choice` to something more specific, like `defaultTab` and `choiceTab` to better convey that they are tab elements and go inside a tab object.

Something like `sage-pane-controller` does get us into a more generic naming. Interesting article here with good guidelines about approaching the `tablist` from an ARIA perspective. That could be a possible name as well. https://www.accessibility-developer-guide.com/examples/widgets/tablists/#:~:text=Tablists%20help%20to%20split%20up,page%20fragments%20inside%20a%20page.

I'll get those spacing issue worked out; thank you!

lib/sage-frontend/stylesheets/system/patterns/elements/_choice.scss
joe-dakroub: This is a good thought about detecting the tab type and adjusting the active class. I can address that issue.

joe-dakroub: To address these issues, I added `data-sage-active-class` to the tab and choice elements. The JS script now uses this attribute to determine which active class to use. I made modifiers within each stylesheet to address the BEM leaks.

app/views/examples/objects/tabs/_props.html.erb
joe-dakroub: The style modifier exists to determine which element type to render, `tab` or `choice`

app/views/examples/objects/tabs/_markup.html.erb
joe-dakroub: I worked out the spacing issues between tab and choice.

lib/sage-frontend/stylesheets/system/patterns/elements/_choice.scss
joe-dakroub: I moved the layout classes into tabs.scss and used a selector to target the child elements to adjust the spacing we need. The removes the BEM leak and puts the control back into the tabs object.

app/views/examples/objects/tabs/_markup.html.erb
philschanely: Validated the spacing change.

philschanely:
In Andrew's absence I'm vouching this is good to merge. We can revisit any potential renaming later as needed.

PR #375, opened 2020-08-11
Add styles to support upload cards

philschanely:
## Description

In preparation to support upload cards this PR adds the following styles to Cards:

- Flat border style with coloring for default and danger
- Dashed border style with coloring for default and danger
- Interactive dashed border style with coloring for default and danger

Screen Shot 2020-08-12 at 11 43 41 AM

Systematizing the design further the following additional changes were made:

- Added a spacer utility
  - Add `sage-spacer-{top|right|bottom|left}-{3xs|2xs|xs|sm|md|lg|xl|2xl|3xl}` to an element to add margin of the specified size on the specified size of the object. NOTE: This is will likely conflict with other margin settings in place and should be used sparingly in lieu of spacing tailored for a particular object.
- Updates labels to use latest color specs from Figma and improves the color combinations variable for wider use
- Added an `icon-card` element for displaying icons in a background box
  - Allows for same color set as labels (uses new color combination variable)
  - Allows for draft as well as bold variation
- Refactored the button partial a smidge


lib/sage-frontend/stylesheets/system/patterns/objects/_card.scss
```diff
@@ -5,11 +5,38 @@
 
 ================================================== */
 
+@mixin sage-card-dashed-border($thickness: 1, $color: sage-color(grey, 400), $dash: 4.5, $gap: 4.5, $offset: 4.5, $border-radius: 4) {
+  $stripped-color-value: str-replace("#{$color}", "#");
+  background-image: url("data:image/svg+xml,%3csvg width='100%25' height='100%25' xmlns='http://www.w3.org/2000/svg'%3e%3crect width='100%25' height='100%25' fill='none' rx='#{$border-radius}' ry='#{$border-radius}' stroke='%23#{$stripped-color-value}' stroke-width='#{$thickness * 2}' stroke-dasharray='#{$dash}%2c#{$gap + $thickness * 4}' stroke-dashoffset='#{$offset}' stroke-linecap='square'/%3e%3c/svg%3e");
```
teenwolfblitzer: It looks like the `width` and `height` values being declared twice here (before and after the `xmlns` declaration)

lib/sage-frontend/stylesheets/system/core/_icons.scss
```diff
@@ -78,9 +78,15 @@ $sage-icon-li-margin-right: 0.4em !default;
       $suffix: "";
     }
 
-    .sage-icon-#{$icon-name}#{$suffix}::before {
-      @include icon-base($icon-name, $size-name);
-      @extend %sage-icon-extensions;
+    .sage-icon-#{$icon-name}#{$suffix} {
+      &::before {
+        @include icon-base($icon-name, $size-name);
+        // @extend %sage-icon-extensions;
```
teenwolfblitzer: Are these comments cleared to be removed?

lib/sage-frontend/stylesheets/system/core/_variables.scss
```diff
@@ -57,6 +57,81 @@ $sage-grid-breakpoint-lg: sage-breakpoint(lg-min) !default;
 $sage-grid-breakpoint-lg-max: sage-breakpoint(lg-max) !default;
 $sage-grid-breakpoint-xl: sage-breakpoint(xl-min) !default;
 
+// ==================================================
+// COLOR COMBINATIONS
+// ==================================================
+$sage-color-combinations: (
```
teenwolfblitzer: Thoughts on separating this piece into a new token? It appears to be a fairly common pattern at this point.

lib/sage-frontend/stylesheets/system/core/_functions.scss
```diff
@@ -53,3 +53,20 @@ $default-baseline: rem(4px) !default;
...
+/// @author Hugo Giraudel
+/// @param {String} $string - Initial string
+/// @param {String} $search - Substring to replace
+/// @param {String} $replace ('') - New value
+/// @return {String} - Updated string
+@function str-replace($string, $search, $replace: "") {
+  $index: str-index($string, $search);
+
+  @if $index {
+    @return str-slice($string, 1, $index - 1) + $replace + str-replace(str-slice($string, $index + str-length($search)), $search, $replace);
```
teenwolfblitzer, line 68: 🧙

teenwolfblitzer (approved):
👍 🔥 Looks great, awesome work! A few items to consider, but nothing blocking here.

lib/sage-frontend/stylesheets/system/core/_icons.scss
```diff
@@ -78,9 +78,15 @@ $sage-icon-li-margin-right: 0.4em !default;
...
+      }
+      &[class*="sage-spacer-top"]::after {
+        @include icon-base($icon-name, $size-name);
+        // @extend %sage-icon-extensions;
```
pixelflips: Can the commented line be removed?

lib/sage-frontend/stylesheets/system/core/_icons.scss
pixelflips: Can the commented line be removed?

pixelflips (approved):
👍  A couple of minor/non-blocking comments, but this is great! Nice work!

lib/sage-frontend/stylesheets/system/core/_variables.scss
philschanely: Great idea! I'm on it.

lib/sage-frontend/stylesheets/system/patterns/objects/_card.scss
philschanely: 🤔  Good point... I was thinking it was once on the container to ensure the svg itself spans the full size of the containing element and then again on the rectangle inside the SVG so that it spans the full size of the SVG... does that clear the concern or can it be simplified further? @monkeypox8

lib/sage-frontend/stylesheets/system/core/_icons.scss
philschanely: Good point. @pixelflips do you or @monkeypox8 know of any place you are using these extensions? I found it was significantly slowing down the compilation on save.

pixelflips: > Good point. @pixelflips do you or @monkeypox8 know of any place you are using these extensions? I found it was significantly slowing down the compilation on save.

@philschanely not that I can think of.

lib/sage-frontend/stylesheets/system/patterns/objects/_card.scss
teenwolfblitzer: Oh, I see! I missed the `rect` in there, my mistake. I also just spotted an extra escaped `%23` (#) for the stroke value.

lib/sage-frontend/stylesheets/system/core/_icons.scss
teenwolfblitzer: Nothing I'm aware of either.

lib/sage-frontend/stylesheets/system/patterns/objects/_card.scss
philschanely: Funny you mention that! I had to puzzle over this a good bit at first. I found that the `%23` is needed there as an encoded symbol instead of the `#` that we usually attach to a hash. So on the prior line I use the new `str-replace()` function to strip that off of the provided color. So `$stripped-color-value` is the hex color value without the `#`. Come to thing of it, I suppose I could help this by replacing `#` with `%23`.

lib/sage-frontend/stylesheets/system/core/_icons.scss
philschanely: OK. I'll pull them for realio then. I believe these are a remnant from converting the old Nucleo export stuff to our system.

joe-dakroub (commented):
Strong update here, Phil! Nice idea about the spacer utility functions. 👍

PR #377, opened 2020-08-12
Downgrade spring to match Product's version

AndrwM:
## Description
```
/Users/andrewmcintee/.gem/ruby/2.4.9/gems/bundler-1.17.3/lib/bundler/runtime.rb:319:in `check_for_activated_spec!': You have already activated spring 2.1.0, but your Gemfile requires spring 1.3.3. Prepending `bundle exec` to your command may solve this. (Gem::LoadError)
```
Happens pretty frequently while in Kajabi-Products. Spring gets stuck running and has to been manually quit/uninstalled so that Product's Spring can run. `bundle exec` doesn't help as the error message suggests.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -118,7 +118,7 @@ GEM
       sprockets (>= 2.8, < 4.0)
       sprockets-rails (>= 2.0, < 4.0)
       tilt (>= 1.1, < 3)
-    spring (2.1.0)
+    spring (1.3.6)
     sprockets (3.7.2)
       concurrent-ruby (~> 1.0)
       rack (> 1, < 3)
@@ -154,7 +154,7 @@ DEPENDENCIES
   rails (= 4.2.11.1)
   sage!
   sass-rails (~> 5.0)
-  spring
+  spring (~> 1.3.3)
   uglifier (>= 1.3.0)
   web-console (~> 2.0)
   webpacker (~> 4.x)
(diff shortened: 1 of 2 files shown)


pixelflips (approved):
👍  +1

philschanely (approved):
Nice! 🏆

Gemfile.lock
```diff
@@ -118,7 +118,7 @@ GEM
       sprockets (>= 2.8, < 4.0)
       sprockets-rails (>= 2.0, < 4.0)
       tilt (>= 1.1, < 3)
-    spring (2.1.0)
+    spring (1.3.6)
```
philschanely, line 121: @AndrwM Just looking back at this as I ran into that issue again but this time arguing about 1.3.6 versus 1.3.3. Is this particular version needed here? May be that I won't encounter the issue again but just though I'd ask.

AndrwM: @philschanely Yeah I ran into that too. I pushed a commit changing that directly to master about 5min after landing this branch when I realized the mistake. A pull from the most recent `master` should fix

AndrwM: The gemfile fuzzy-matched to 1.3.6 instead of 1.3.3 like I wanted, my bad on that :)

philschanely: No problem. Thanks!!

PR #380, opened 2020-08-12
Revise sage type implementation

philschanely:
## Description
This PR pushes more on the spacing issues we've encountered in the docs site when `sage-type` is added on a container. Theory is that this can be revised to affect only direct children.

Also refactors a bit to make a new token file isolating the type specs there. I do a smidge of variable refactoring to reduce/move global type variables.

### DIFF
--- a/app/views/examples/elements/button/_preview.html.erb
+++ b/app/views/examples/elements/button/_preview.html.erb
@@ -52,7 +52,7 @@
   text: "Button (icon-right)"
 %>
 
-<br><br><br>
+
 <h3 class="t-sage-heading-6">Primary (disabled)</h3>
 
 <!-- Primary button (disabled) -->
@@ -107,7 +107,7 @@
   text: "Link (icon-right)"
 %>
 
-<br><br><br>
+
 <h3 class="t-sage-heading-6">Secondary</h3>
 
 <!-- Secondary button -->
@@ -162,7 +162,7 @@
   text: "Link (icon-right)"
 %>
 
-<br><br><br>
+
 <h3 class="t-sage-heading-6">Secondary (disabled)</h3>
 
 <!-- Secondary button (disabled) -->
@@ -217,7 +217,7 @@
   text: "Button (icon-right)"
 %>
 
-<br><br><br>
+
 <h3 class="t-sage-heading-6">Tertiary</h3>
 
 <!-- Tertiary button -->
@@ -272,7 +272,7 @@
   text: "Button (icon-right)"
 %>
 
-<br><br><br>
+
 <h3 class="t-sage-heading-6">Tertiary (disabled)</h3>
 
 <!-- Tertiary button disabled -->
@@ -327,7 +327,7 @@
   text: "Link (icon-right)"
 %>
 
-<br><br><br>
+
 <h3 class="t-sage-heading-6">Danger</h3>
 
 <!-- Danger button -->
@@ -383,7 +383,7 @@
 %>
 
 
-<br><br><br>
+
 <h3 class="t-sage-heading-6">Danger (disabled)</h3>
 
 <!-- Danger button (disabled) -->
@@ -426,7 +426,7 @@
 %>
 
 
-<br><br><br>
+
 <h3 class="t-sage-heading-6">Flex Align End</h3>
 
 <!-- Primary button (icon-right) -->
@@ -456,7 +456,7 @@
 %>
 
 
-<br><br><br>
+
 <h3 class="t-sage-heading-6">Small</h3>
 
 <!-- Primary small -->
@@ -499,7 +499,7 @@
 %>
 
 
-<br><br><br>
+
 <h3 class="t-sage-heading-6">Tiny Button (Limited use)</h3>
 
 <button class="sage-btn sage-btn--tertiary sage-btn--icon-left-tag sage-btn--tiny">
@@ -510,7 +510,7 @@
 </button>
 
 
-<br><br><br>
+
 <h3 class="t-sage-heading-6">Icon Button</h3>
 
 <strong class="t-sage-body-small-med">Regular</strong>
@@ -566,7 +566,7 @@
   text: "Name Of Action"
 %>
 
-<br><br>
+
 <strong class="t-sage-body-small-med">Small</strong>
 <!-- Primary button (icon-right) -->
 <%= render "examples/elements/button/markup",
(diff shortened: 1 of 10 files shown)


lib/sage-frontend/stylesheets/system/core/_typography.scss
```diff
@@ -138,114 +25,122 @@ $type-specs: (
...
-  ol:not([class]),
-  dl:not([class]),
-  [class*="t-sage-body"] {
-    @include t-block-spacing($sage-body-margin-bottom);
+  > p:not([class]),
+  > ul:not([class]),
+  > ol:not([class]),
+  > dl:not([class]),
+  > [class*="t-sage-body"] {
+    @include t-block-spacing($-body-margin-bottom);
```
AndrwM, line 57: Nice!

lib/sage-frontend/stylesheets/system/core/_variables.scss
```diff
@@ -22,20 +22,12 @@ $sage-transition: 0.1s ease-in-out !default;
...
 
 // Text
-$sage-body-font-size: 1rem !default; // 16px
-$sage-body-font-weight: sage-font-weight() !default;
-$sage-body-font-line-height: sage-line-height(md) !default;
+$sage-body-font-size: rem(16px) !default;
 $sage-body-font-color: sage-color(charcoal, 400) !default;
 $sage-body-letter-spacing: sage-letter-spacing(xs) !default;
-$sage-body-margin-bottom: sage-spacing(sm) !default;
+// -----======|||0|||======------ //
```
AndrwM: Haha, nice divider

AndrwM (approved):
Great work! 🚀

lib/sage-frontend/stylesheets/system/core/_variables.scss
philschanely: Thanks! I should probably pull it though ;)

lib/sage-frontend/stylesheets/docs/_variables.scss
```diff
@@ -12,6 +12,7 @@
 //   --something: something;
 // }
 
+$docs-font-monospace: "SFMono-Regular", "Menlo", "Monaco", "Consolas", "Liberation Mono", "Courier New", monospace !default;
```
teenwolfblitzer, line 15: Curious if we have a font stack preference from design when it comes to monospace fonts. Not that I'd want to suggest adding another font asset just for the documentation, but I spy an instance or two of Roboto Mono in Figma.

lib/sage-frontend/stylesheets/system/core/_typography.scss
teenwolfblitzer: 🚀

lib/sage-frontend/stylesheets/system/tokens/_type_specs.scss
```diff
@@ -0,0 +1,112 @@
...
+  size: sage-font-size(sm),
+  leading: sage-line-height(sm),
+));
+
+$-t-body-xsmall-spec: map-merge($-t-base-spec, (
+  size: sage-font-size(xs),
+  leading: sage-line-height(xs),
+));
+
+// The following map is used to generate both `%t-` classeses for each entry.
```
teenwolfblitzer: Teeny tiny typo here, `classeses` instead of `classes`

app/views/examples/elements/button/_preview.html.erb
```diff
@@ -52,7 +52,7 @@
   text: "Button (icon-right)"
 %>
 
-<br><br><br>
+
```
teenwolfblitzer, line 55: 🙇

lib/sage-frontend/stylesheets/system/core/_typography.scss
```diff
@@ -265,7 +160,9 @@ hr {
   border-top: sage-border();
 }
 
+// Generat text color modifiers
```
teenwolfblitzer: Missing an `e` in `Generate` here

lib/sage-frontend/stylesheets/system/tokens/_type_specs.scss
```diff
@@ -0,0 +1,112 @@
...
+  "body-xsmall-bold": map-merge($-t-body-xsmall-spec, (
+    weight: sage-font-weight(bold)
+  )),
+
+  // Additional custom specs
+);
+
+// Set up base specs as extendables
+@each $spec-name, $spec-map in $sage-type-specs {
+  %t-sage-#{$spec-name} {
```
teenwolfblitzer, line 123: Thoughts on outputting this function as a `mixin` rather than an `extend` (or possibly as both, to avoid potential conflicts)? I've found in more than one situation that combined classes using extends get "hoisted" in the order, negating BEM and/or specificity in reused styles.

teenwolfblitzer (approved):
👍    Looks great, much much cleaner than before. Great work! A few items to check out, nothing blocking.

lib/sage-frontend/stylesheets/system/tokens/_type_specs.scss
philschanely: Interesting! Sure. I can certainly make a type spec mixin and use it in this loop to generate the extends. However, I am a little concerned about the potential confusion/mix of implementing `mixin` sometimes and `extend` other times. While I have found that I like seeing the grouped selector the `extend` creates that thus separates the type styles form the others in the inspector, do you think `mixin` would be a safer one-size-fits-all solution over `extend`? @monkeypox8 

CC: @Kajabi/ux-devs

lib/sage-frontend/stylesheets/docs/_variables.scss
philschanely: 🤔

lib/sage-frontend/stylesheets/system/tokens/_type_specs.scss
teenwolfblitzer: Agreed, benefits to both approaches but we should stick to one to avoid confusion. My preference leans towards using `mixin` over `extend` globally, since it doesn't reorder CSS once compiled. For a sample of what I was seeing, see the second example under "About those placeholders": https://css-tricks.com/sass-techniques-from-the-trenches/#about-those-placeholders. Now I'm wondering if anyone else has run into similar issues within the existing app structure?

In any case, this can definitely be for a follow-up discussion with the group, no need to hold up this PR!

philschanely: Sounds good. I'll push this one in but we do now have the mixin so refactoring should be an easy follow-up.

PR #382, opened 2020-08-15
Sync Radio buttons with spec 

teenwolfblitzer:
## Description
Radio button size no longer matches the Figma file. We have a `24px` size radio button, while the spec is now calling for `16px`.

|  Expected   |  Actual  |
|--------|--------|
|||

## Related

- n/a

### DIFF
--- a/lib/sage-frontend/stylesheets/system/core/_variables.scss
+++ b/lib/sage-frontend/stylesheets/system/core/_variables.scss
@@ -213,10 +213,10 @@ $sage-radio-color-hover: sage-color(grey, 500) !default;
 $sage-radio-color-disabled: sage-color(grey, 100) !default;
 $sage-radio-color-disabled-checked: sage-color(grey, 300) !default;
 
-$sage-radio-button-size: rem(24px) !default;
+$sage-radio-button-size: rem(16px) !default;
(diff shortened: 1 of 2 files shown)


lib/sage-frontend/stylesheets/system/core/_variables.scss
```diff
@@ -213,10 +213,10 @@ $sage-radio-color-hover: sage-color(grey, 500) !default;
 $sage-radio-color-disabled: sage-color(grey, 100) !default;
 $sage-radio-color-disabled-checked: sage-color(grey, 300) !default;
 
-$sage-radio-button-size: rem(24px) !default;
+$sage-radio-button-size: rem(16px) !default;
 $sage-radio-border-radius: sage-border(radius-round) !default;
 $sage-radio-transition: 0.15s ease-in-out !default;
-$sage-radio-selected-indicator-size: rem(16px) !default;
+$sage-radio-selected-indicator-size: rem(7px) !default;
```
teenwolfblitzer, line 219: Spec calls for `6.67px` but using the fractional value appears to introduce some rounding error/ horizontal misalignment when applied with the transform.

pixelflips (approved):
👍 +1 Thanks for updating!

PR #384, opened 2020-08-17
Sage Lists

philschanely:
## Description

This PR adds a class and modifiers for unstyled and inline lists.

Screen Shot 2020-08-18 at 11 49 45 AM


pixelflips (commented):
@philschanely Looks great, I did notice a couple of things.

1. The tiles don't wrap on smaller screens (may not be a requirement for the PR)
2. There seems to be a lint failure. (missing newline)

philschanely:
> @philschanely Looks great, I did notice a couple of things.
> 
> 1. The tiles don't wrap on smaller screens (may not be a requirement for the PR)
> 2. There seems to be a lint failure. (missing newline)

Thanks, @pixelflips . Fixed the linting. You're right about the wrapping and I believe it is something we should look at closer in future as need for this is clarified over time. Perhaps I'm even overthinking it in this PR 🤔 . I will request a review after looking at this a little more.

app/views/examples/elements/tiles/_preview.html.erb
```diff
@@ -0,0 +1,119 @@
...
+  )
+%>
+
+<h5>Checkbox list</h5>
+<%= render "examples/elements/tiles/markup", 
+  layout: "inline-4",
+  items: %(
+    <li class="sage-checkbox">
+      <input class="sage-checkbox__input" id="check-0" name="group-2" type="checkbox" value="option-1">
+      <label class="sage-checkbox__label" for="check-0">It matters a whole lot.</label>
```
teenwolfblitzer: Looks like the IDs for each group are getting reused/repeated in the preview here, so the labels aren't applying to their respective controls.

lib/sage-frontend/stylesheets/system/patterns/elements/_tiles.scss
```diff
@@ -0,0 +1,40 @@
...
+      margin-left: sage-spacing(sm);
+    }
+  }
+}
+
+.sage-tiles--inline {
+  flex-flow: row wrap;
+}
+
+@for $i from 2 through 5 {
```
teenwolfblitzer: Should the limit of 5 tile elements be noted in the documentation?

teenwolfblitzer (commented):
Looking great, one item to look into with the IDs being repeated, should be good to go once that's sorted.

Also, I noticed there's a negative margin being applied on sibling switch elements, presumably to reduce vertical spacing. It gets a bit weird when switches are placed horizontally in the list format:

sage-switch-row

philschanely:
@monkeypox8 and @pixelflips Thanks for your review of the earlier rendition of this PR. I have decided on a different tack of just providing and unstyled list option rather than overthinking this with different layouts as tiles. Hopefully this fits what we need a little more since the grid helpers provide a lot more flexibility for responsive styling structured layouts.

app/views/examples/elements/lists/_preview.html.erb
```diff
@@ -0,0 +1,86 @@
...
+  <li class="sage-switch">
+    <input class="sage-switch__input" id="switch-3a" name="group-1" type="radio" value="option-4">
+    <label class="sage-switch__label" for="switch-3a">Undecided.</label>
+  </li>
+</ul>
+
+<h5>Inline List</h5>
+<p>Inline lists will wrap by default so this works best with items that have an intrinsically limited smaller width that affords space for other elements to a point.</p>
+<ul class="sage-list sage-list--inline">
+  <li class="sage-switch">
```
pixelflips, line 61: This one is minor, but the first list item doesn't quite align with the others. From what I can tell there may be a negative margin applied to the others. See `_switch.scss `around line 15.

app/views/examples/elements/lists/_props.html.erb
```diff
@@ -0,0 +1,30 @@
...
+    Note that these three booleans are set up with stacked modifiers:
+    <ul>
+      <li>Default -- <code>sage-list</code></li>
+      <li>Inline, wrapping, spaced -- <code>sage-list--inline</code></li>
+      <li>Inline, wrapping, compact -- <code>sage-list--inline-compact</code></li>
+      <li>Inline, non-wrapping, spaced -- <code>sage-list--inline-fit</code></li>
+      <li>Inline, non-wrapping, compact -- <code>sage-list--inline-fit-compact</code></li>
+    </ul>
+  </td>
+</tr>
```
pixelflips, line 30: Not blocking at all, but a few of the files are missing the newline at the end.

pixelflips (approved):
Just a couple of minor issues, but nothing major or blocking. Nice work!

app/views/examples/elements/lists/_preview.html.erb
philschanely: Thanks, @pixelflips . I have addressed this and lined up the other serial fields with a similar fix.

PR #386, opened 2020-08-18
[BUG] Code status page displays an error

teenwolfblitzer:
## Description
[Code status page](https://sage-design-system.kajabi.com/pages/status) is not rendering when viewed. 

Issue seems to be related to some of the generated objects missing `docs` in their definition(s).


philschanely (approved):
Thanks, @monkeypox8 !!

PR #333, opened 2020-08-18
Update quick link pseudo-states

pixelflips:
## Description

Match quick link `hover` `focus` and `active` states to revised sidebar nav styles.

## Related

- #329


philschanely (approved):
Looks good!

PR #388, opened 2020-08-20
Update containers and Cards

philschanely:
## Description

This PR makes some subtle updates to containers and Cards:

- Revises the nomenclature for containers so that BEM is implemented correctly: `sage-container` on its own defaults to a 100% fluid container with side padding. Other size values such as `sm`, `md`, etc. can be added as a modifier to set a particular maximum width to the container. This change is worked through the documentation markup where `container` was implement and will need to be updated in `products` as well.
- Revises approach to the `compact` Card variation to use CSS Grid. This results in the footer and body being more tidily stacked atop each other while both sit beside the image.
- Revises approach to the image element on Cards so that the image is moved inside a `div` and set up to use `object-fit: cover` so that the image is more readily responsive. This revision will need to be made in `kajabi-products` as well.

On the side a few updates were made to the documentation:

- Updates documentation for containers to demonstrate them more clearly.
- Sandbox page is made a more wide-open for experimentation by removing sidebars and adjusting the containers used.

Lastly, I patched a few little things related to radio and checkbox sizes and spacings.


app/views/examples/objects/icon_list/_preview.html.erb
```diff
@@ -42,13 +42,13 @@
       ),
       bullet_type: "checkbox",
       body: %(
-        <label for="c1" class="t-sage-body-med">Grant an offer</label>
-        <p class="t-sage-body-small t-sage--color-charcoal-200">
+        <label for="c1" class="sage-icon-list-item__label">Grant an offer</label>
```
joe-dakroub, line 45: For my clarification, the label has `for="c1"`. What element does this reference?

joe-dakroub (approved):
One minor question, not a blocker, but really nice cleanup and simplification here Phil. Thanks! 👍

app/views/examples/objects/icon_list/_preview.html.erb
philschanely: Its for the accompanying input a few lines up I believe.

philschanely: Its for the accompanying input a few lines up I believe.

PR #390, opened 2020-08-20
Modal Bugfix: Safari Animation

AndrwM:
## Description
In Safari, the Sage Modal, when using the `transition` property:
The browser intermittently hid the modal and/or UI flashes would occur. I believe this has to do with `transition` -ing a property as a child of a parent using the `backdrop-filter` property.


philschanely (approved):
Yes, we used to run into this a lot on a previous project. It might be worth trying to activate browser hardware acceleration by setting something like `transform: scale3d(0,0,0,)`. Similar concern discussed here: https://medium.com/@daoanhvu/scrolling-performance-hack-with-chrome-dev-tool-19c264523289

AndrwM:
Thanks @philschanely, I'll take a look 🧑🏻‍💻

PR #391, opened 2020-08-24
Bugfix: modal blur does not dismiss

AndrwM:
## Description
Ripping all animation related to SageModals. Its causing some UI weirdness in Chrome when using `backdrop-filter`.
The current animation effect is nominal since our animation only runs for 0.1s for SageModal animations.
Additionally this simplifies the CSS governing our SageModals

## Related
Issue reported here:
https://kajabi.slack.com/archives/CG9MC1SKT/p1598051302008500


philschanely (approved):
LGTM 👍

PR #392, opened 2020-08-25
Jcm remove utilities

cmcfadzean:
## Description
The only utility that we were showing was a media that is not used. This removes that and takes the page off the sidebar. If we feel that we are missing it we can always add it back as the code to show utilities is still there.


philschanely (approved):
I'm fine with this given how things are now. I think there may be some need for documenting utilities we are using after we audit/if we still find them necessary. 🏅

PR #393, opened 2020-08-25
Panel adjustments

philschanely:
## Description

This PR adds some improvements to the Panel styles according to specs:

- Allows for a panel-group to be labeled but not inline unless also inside a scroll-container
- Allows for panel-group to have a color modifier that affects contents the same way a color modifier on the panel itself does
- Improves color inheritance to override Ladera defaults
- Fixes spacing below inline panels
- Adds a truncate utility class `t-sage--truncate` to make single-line text with `...` when no room remains in its container.

NOTE: Accompanying adjustments will need to be made in `products`.

### DIFF
--- a/lib/sage-frontend/stylesheets/system/core/_mixins.scss
+++ b/lib/sage-frontend/stylesheets/system/core/_mixins.scss
@@ -254,3 +254,15 @@
   $stripped-color-value: str-replace("#{$color}", "#", "%23");
   background-image: url("data:image/svg+xml,%3csvg width='100%25' height='100%25' xmlns='http://www.w3.org/2000/svg'%3e%3crect width='100%25' height='100%25' fill='none' rx='#{$border-radius}' ry='#{$border-radius}' stroke='#{$stripped-color-value}' stroke-width='#{$thickness * 2}' stroke-dasharray='#{$dash}%2c#{$gap + $thickness * 4}' stroke-dashoffset='#{$offset}' stroke-linecap='square'/%3e%3c/svg%3e");
 }
+
+/* ==================================================
+  ** Truncate to a single line
+  Usage: `@include sage-truncate;`
+
+  Truncates text of the given element to a single line.
+================================================== */
+@mixin sage-truncate() {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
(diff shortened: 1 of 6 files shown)


lib/sage-frontend/stylesheets/system/core/_mixins.scss
```diff
@@ -254,3 +254,15 @@
...
+  ** Truncate to a single line
+  Usage: `@include sage-truncate;`
+
+  Truncates text of the given element to a single line.
+================================================== */
+@mixin sage-truncate() {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
```
AndrwM, line 268: Love this util, nice

lib/sage-frontend/stylesheets/system/patterns/objects/_panel.scss
```diff
@@ -34,17 +33,14 @@ $-panel-spacing-sm: sage-spacing(xs);
...
-  }
-
-  .sage-panel-group__items &:not(:last-child) {
-    margin-right: sage-spacing(sm);
-  }
-
   .sage-panel-scroll-container & {
-    min-width: 180px;
+    min-width: 160px;
+    max-width: 200px;
```
AndrwM: This sizing seems super specific. I'm not familiar with how this scrolling pattern is used but if this is a one-off layout for a view I think this would be best implemented in Products rather than the design system.

lib/sage-frontend/stylesheets/system/patterns/objects/_panel.scss
```diff
@@ -134,7 +135,8 @@ $-panel-spacing-sm: sage-spacing(xs);
     }
 
     @else {
-      .sage-panel--#{$-color-name} & {
+      .sage-panel--#{$-color-name} &,
+      .sage-panel-gropu--#{$-color-name} & {
```
AndrwM: Spelling

AndrwM (approved):
LGTM! Spotted a spelling error. These panels are quickly becoming multi-varianted and I wonder if we should start discussions of splitting these into multiple components. (Dividing SageAccordion & SagePanel comes to mind initially)

lib/sage-frontend/stylesheets/system/patterns/objects/_panel.scss
philschanely: Good call. I debated that myself too and hearing you voice that I will pull it out.￼

lib/sage-frontend/stylesheets/system/core/_mixins.scss
joe-dakroub: I second that!

joe-dakroub (approved):
Looks good Phil. Nice updates per usual. 👍

PR #394, opened 2020-08-25
sidebar order

cmcfadzean:
## Description
The sidebar is getting a bit cluttered with some things that are linking to the same places. This reduces the size and moves the changelog to the top next to the current version info.


pixelflips (approved):
Nice! 👍 That's a great place for it!

PR #396, opened 2020-08-25
[BUG] Index image breaks out of container on smaller screens

pixelflips:
## Description
The image on the index page breaks out of its container on smaller screens.

### Screenshots
|  Before   |  After  |
|--------|--------|
|localhost_4000_pages_index|localhost_4000_pages_index (1)| 

### Steps for testing
* Open Foundations page
* Resize browser
* Verify image at the bottom of the page no longer escapes container.

## Related issues
- n/a

### DIFF
--- a/app/views/pages/index.html.erb
+++ b/app/views/pages/index.html.erb
@@ -12,7 +12,7 @@
       <p>We think of our approach to UI at Kajabi in this way: We default to using Rails Components and a classic Rails approach to most problems and move to React where it counts.</p>
       <p>Rails does many things very well and gives us the ability to move quickly and solve complex problems with very simple, tried and true, code solutions. With that said, sometimes we come across a problem that we want to solve that the standard rails approach will just not be enough ( Think complex interactions with many versions of state ). In this case we move to React as our default approach to solving complex, Javascript heavy problems.</p>
       <p>Because our system contains two different approaches to UI creation, we utilize a SCSS Design System to provide the styles to both types of components (Think Bootstrap, but customized for our products). Our Design System provides the styles for the core components that make up the UI of our product.</p>
-      <%= image_pack_tag("docs/sage_structure.png", class: "sage-panel__img", style: "display: block; max-width: 500px; margin: 0 auto") %>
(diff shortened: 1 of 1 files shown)


app/views/pages/index.html.erb
```diff
@@ -12,7 +12,7 @@
...
       <p>Rails does many things very well and gives us the ability to move quickly and solve complex problems with very simple, tried and true, code solutions. With that said, sometimes we come across a problem that we want to solve that the standard rails approach will just not be enough ( Think complex interactions with many versions of state ). In this case we move to React as our default approach to solving complex, Javascript heavy problems.</p>
       <p>Because our system contains two different approaches to UI creation, we utilize a SCSS Design System to provide the styles to both types of components (Think Bootstrap, but customized for our products). Our Design System provides the styles for the core components that make up the UI of our product.</p>
-      <%= image_pack_tag("docs/sage_structure.png", class: "sage-panel__img", style: "display: block; max-width: 500px; margin: 0 auto") %>
+      <%= image_pack_tag("docs/sage_structure.png", class: "sage-panel__img", style: "display: block; max-width: 500px; margin: 0 auto; width: 100%") %>
```
teenwolfblitzer: Should we apply the anti-jank sizing [now in Chrome/Firefox/Safari](https://css-tricks.com/do-this-to-improve-image-loading-on-your-website/) by setting unitless `width` and/or `height` values here rather than as inline styles?

I also considered recommending `img` and `svg` be set globally to `max-width: 100%` and `height: auto` to keep this from happening again, but since our use of images in the admin is much more measured vs a typical website, it's probably overkill.

teenwolfblitzer (approved):
👍 LGTM, I think this works fine for this instance. Left a suggestion below to consider.

app/views/pages/index.html.erb
pixelflips: @monkeypox8 yeah, I thought about the `max-width: 100%` and `height: auto` as well, but came to the same conclusion. Good call on the anti-jank 🙌  - will add that as well!

PR #397, opened 2020-08-26
Updates to Dropdown

philschanely:
## Description

- Adds a WIP Search element and a search item to dropdowns.
- Revises dropdown js to ignore clicks on search input fields.

Screen Shot 2020-08-27 at 7 39 59 AM


lib/sage-frontend/stylesheets/system/patterns/objects/_dropdown.scss
```diff
@@ -143,8 +142,21 @@ $option-icon: "::before";
...
+  width: 100%;
+}
+
+.sage-dropdown__trigger--dropdown,
+.sage-dropdown__trigger--field {
+  justify-content: space-between;
+}
+
+.sage-dropdown__trigger--dropdown {
+  min-width: 280px;
```
joe-dakroub, line 161: Is this a standard measurement in Sage or something we should use a variable for?

joe-dakroub (approved):
Looking good locally. I left one comment, not a blocker. Nice work Phil! 👍

lib/sage-frontend/stylesheets/system/patterns/objects/_dropdown.scss
philschanely: Good catch! I should at least make it a variable and use the `rem()` function. Will address.￼

PR #398, opened 2020-08-27
Spdate sage-local-link documentation

AndrwM:
## Description
- I noticed that the usage logging help wasn't clear, updated.
- Added a `sage-local-link.sh status` option to check if there's a symlinked Sage. I find myself often running the setup/teardown just to double check which Sage instance I'm using.
- Better file commenting


joe-dakroub (approved):
Nice update Andrew. Thanks for the help messages! 👍

PR #410, opened 2020-09-09
Add icons to labels

philschanely:
## Description

Add icons to labels.

Screen Shot 2020-09-09 at 3 20 43 PM


joe-dakroub (approved):
Looking good; nice new options! 👍

PR #413, opened 2020-09-15
Update page heading to include additional items

pixelflips:
## Description
Updates page headings to include action items, intro text, toolbar, and secondary text.

### Screenshots

|  Before  |  After  |
|--------|--------|
|Screen Shot 2020-09-15 at 1 25 16 PM|Screen Shot 2020-09-22 at 4 21 05 PM|

## Related
- NA


pixelflips:
@AndrwM this is the initial draft for the page heading updates. Feel free to pull it down and adjust however you see fit and also add the additional variation we discussed. Thanks in advance for the help and guidance!

app/views/examples/objects/page_heading/_preview.html.erb
```diff
@@ -1,6 +1,137 @@
...
-  page_title: "Page Title",
-  back_icon: "sage-icon-caret-left",
+<% preview_button = sage_component(SageButton,{style: "tertiary",icon:{name: "add",style: "left"},value: "Preview",}) %>
+<% create_button = sage_component(SageButton,{style: "primary",icon:{name: "add",style: "left"},value: "Create",}) %>
+
+<h3 class="t-sage-heading-6">Standalone Page Heading</h3>
+
+<%= sage_component SagePageHeading, {
+  title: "Page Title",
+  title_copy: "",
```
philschanely: Could we do something to make these props not required and thus trim down the space in our preview file? Not sure how to do this with the components but before I would do something like `if defined?(prop)` in the template file. I imagine you know more about this than I do though ;)

lib/sage-frontend/stylesheets/system/patterns/objects/_page_heading.scss
```diff
@@ -5,18 +5,80 @@
 ================================================== */
 
 .sage-page-heading {
-  display: flex;
-  flex-flow: column;
-  margin-bottom: sage-spacing(lg);
+  display: grid;
+  grid-template-areas:
+    "crumbs crumbs"
```
philschanely: FYI if there's a chance you don't want `crumbs` to spill into the space above the `actions` you can use `crumbs . ` to make an empty cell.

lib/sage-frontend/stylesheets/system/patterns/objects/_page_heading.scss
```diff
@@ -5,18 +5,80 @@
...
+  grid-area: toolbar;
+  margin-top: sage-spacing(sm);
+  > * {
+    margin-right: sage-spacing(sm);
+    color: sage-color(charcoal, 200);
+  }
+}
+
+@each $-icon-name, $-icon-code in $sage-icons {
+  .sage-page-heading__toolbar--icon-#{$-icon-name} {
```
philschanely: May be that we'll solve this differently in light of our button discussion. But did you find that this loop slowed down the compiling? I am a little concerned that we not have too many loops like this where we go through our entire icon library. May be that we should start consolidating such things into one place and thus only do the loop once. Just thinking out loud.

lib/sage_rails/app/views/sage_components/_sage_page_heading.html.erb
```diff
@@ -1,23 +1,40 @@
...
-  <% if component.link_url.present? %>
-    <nav aria-label="Breadcrumbs" class="sage-page-heading__back sage-breadcrumbs">
-      <p class="sage-breadcrumbs__item">
-        <a class="sage-breadcrumbs__link sage-breadcrumbs__link--current" aria-current="page" href="<%= component.link_url %>">
-          <i class="sage-breadcrumbs__icon sage-icon-caret-left" aria-label="navigate to"></i>
-          <%= component.link_text %>
-        </a>
-      </p>
-    </nav>
+  <div class="sage-page-heading__title">
```
philschanely: Should the title be first in the source order? I love that Grid gives us the full flexibility to think more about what source order makes the most sense to screen readers. Perhaps there's more to reorder as a result?

lib/sage_rails/app/views/sage_components/_sage_page_heading.html.erb
```diff
@@ -1,23 +1,40 @@
...
+    <h1 class="t-sage-heading-1"><%= component.title %><% if component.help_link.present? %> <a href="<%= component.help_link[:url] %>" class="sage-link--no-launch sage-link--help-icon-only" target="_blank" referrer="no-referrer"><span class="visually-hidden"><%= component.help_link[:label] %></span></a><% end %></h1>
+    <% if component.title_copy.present? %>
+      <p class="sage-page-heading__title--help"><%= component.title_copy %></p>
```
philschanely: This should not be a modifier; should it perhaps be `.sage-page-heading__lead`, or `__copy` or `__intro`?

lib/sage_rails/app/views/sage_components/_sage_page_heading.html.erb
```diff
@@ -1,23 +1,40 @@
...
+      <%= action %>
+    <%- end -%>
+  </div>
   <% end %>
-  <%= component.content %>
-</div>
+  <% if component.toolbar_items.any? %>
+    <div class="sage-page-heading__toolbar">
+    <%- component.toolbar_items.each do |item| -%>
+      <a href="<%= item[:url] %>" class="<%= "sage-page-heading__toolbar--icon-#{item[:icon]}" if defined?(item[:icon]) %>"><%= item[:text] %></a>
```
philschanely: This is also an incorrect application of a modifier. Consider instead `sage-page-heading__toolbar-icon-...`. 

In case its helpful reminder to clarify:

- a modifier should be a style modification to another element or block.
- modifiers are always stacked with the original block or element name. So if a `page-heading` block had a "super" modifier, it would be `page-heading--super` and you'd provide it with `page-heading page-heading--super`. Or if a Page Heading's Title element had a "warning" modifier it would be `page-heading__title--warning` and you'd provide it with `page-heading__title page-heading__title--warning`

You seem to be using it here and in the place above as an additional nested element. Such cases can just be treated just as elements themselves, named to fit whatever the thing semantically is in context. 

Hope this is helpful.

lib/sage-frontend/stylesheets/system/patterns/objects/_page_heading.scss
```diff
@@ -5,18 +5,80 @@
...
+  justify-content: flex-end;
+  margin-top: sage-spacing(sm);
+  color: sage-color(charcoal, 200);
+  font-size: rem(14);
+  font-style: italic;
+}
+
+@media (max-width: sage-breakpoint()) {
+  .sage-page-heading {
+    grid-template-columns: span 2;
```
philschanely: Should probably just do one column here.

lib/sage-frontend/stylesheets/system/patterns/objects/_page_heading.scss
```diff
@@ -5,18 +5,80 @@
...
+  display: flex;
+  grid-area: secondary;
+  justify-content: flex-end;
+  margin-top: sage-spacing(sm);
+  color: sage-color(charcoal, 200);
+  font-size: rem(14);
+  font-style: italic;
+}
+
+@media (max-width: sage-breakpoint()) {
```
philschanely: Let's prefer nesting media queries inside the affected selector rather than grouping them together. This way you can see right inside the selector all the styles that affect it.

lib/sage-frontend/stylesheets/system/patterns/objects/_page_heading.scss
```diff
@@ -5,18 +5,80 @@
 ================================================== */
 
 .sage-page-heading {
-  display: flex;
-  flex-flow: column;
-  margin-bottom: sage-spacing(lg);
+  display: grid;
```
philschanely: If these are not in effect for certain viewports lets guard them with a `min-width` filter. This makes what's happening at each viewport clearer versus what is not affected by viewport.

lib/sage-frontend/stylesheets/system/patterns/objects/_page_heading.scss
```diff
@@ -5,18 +5,80 @@
...
+  grid-template-areas:
+    "crumbs crumbs"
+    "title actions"
+    "toolbar secondary";
+  grid-template-columns: 1fr max-content;
```
philschanely: `1fr auto` may suffice here. Or you might consider a ratio-based `minmax()` to help ensure that a wider right side does not dominate and squash the left side.

lib/sage-frontend/stylesheets/system/patterns/objects/_page_heading.scss
```diff
@@ -5,18 +5,80 @@
...
+    white-space: nowrap;
+    &::before {
+      margin-right: sage-spacing(xs);
+      @include sage-icon-base($-icon-name);
+    }
+  }
+}
+
+.sage-page-heading__actions {
+  display: flex;
```
philschanely, line 54: You could probably add `gap: sage-spacing(sm)` here to achieve the desired space between buttons.

lib/sage-frontend/stylesheets/system/patterns/objects/_page_heading.scss
```diff
@@ -5,18 +5,80 @@
 ================================================== */
 
 .sage-page-heading {
-  display: flex;
-  flex-flow: column;
-  margin-bottom: sage-spacing(lg);
+  display: grid;
+  grid-template-areas:
```
philschanely: I think you can now add `grid-gap: sage-spacing(sm)` to enforce the desired spacing and remove margins on the elements themselves. Another beautiful grid feature! 🙌

lib/sage-frontend/stylesheets/system/patterns/objects/_page_heading.scss
```diff
@@ -5,18 +5,80 @@
...
 }
 
 .sage-page-heading__title {
-  @extend %t-sage-heading-1;
+  grid-area: title;
+}
 
-  order: 2;
+.sage-page-heading__title--help {
+  margin-top: sage-spacing(sm);
```
philschanely, line 26: Depending on what this is supposed to be, it may be more effective to use a grid cell alignment run on the `title` element.

lib/sage-frontend/stylesheets/system/patterns/objects/_page_heading.scss
```diff
@@ -5,18 +5,80 @@
...
+  justify-content: flex-end;
+}
+
+.sage-page-heading__secondary {
+  display: flex;
+  grid-area: secondary;
+  justify-content: flex-end;
+  margin-top: sage-spacing(sm);
+  color: sage-color(charcoal, 200);
+  font-size: rem(14);
```
philschanely: Can we extend a type spec here instead of setting a font size? This appears to be the first time I've seen `italic` used in our app! Consider whether this even deserves a new type spec added.

lib/sage-frontend/stylesheets/system/patterns/objects/_page_heading.scss
philschanely: Looking closer I think this should be on its own row in the grid. Beauty of the grid is that if this is not present then the row flattens to 0. Consider that being:
```
...
intro .
...
```

philschanely (changes requested):

app/views/examples/objects/page_heading/_preview.html.erb
AndrwM: What is the usage difference between `title_copy` and `title`?

lib/sage-frontend/stylesheets/system/patterns/objects/_page_heading.scss
```diff
@@ -5,18 +5,80 @@
...
 }
+
+.sage-page-heading__toolbar {
+  display: flex;
+  grid-area: toolbar;
+  margin-top: sage-spacing(sm);
+  > * {
+    margin-right: sage-spacing(sm);
+    color: sage-color(charcoal, 200);
+  }
```
AndrwM: Color and spacing should be a concern of the child components here. We should probably remove this to prevent unexpected cascade behavior.

lib/sage_rails/app/views/sage_components/_sage_page_heading.html.erb
```diff
@@ -1,23 +1,40 @@
...
-  <% if component.title.present? %>
-    <h1 class="sage-page-heading__title">
-      <%= component.title %>
-      <% if component.help_link.present? %>
-        <a href="<%= component.help_link[:url] %>" class="sage-link--help-icon-only" target="_blank" referrer="no-referrer">
-          <span class="visually-hidden"><%= component.help_link[:label] %></a>
-        </a>
-      <% end %>
-    </h1>
+<section class="sage-page-heading sage-spacer-bottom-sm">
```
AndrwM: If we're including a spacer by default within a component it should be within the `sage-page-heading` scss file rather than being defined as a class.

However if you want to add bottom margin in some instances but not all instances add ```<%= generated_css_classes %>``` to this node will allow devs to change the spacer when needed.

This is what the implementation could look like on the Products side:
```
<%= sage_component SagePageHeading, {
  title: "Page Title",
  spacer: { bottom: :sm },
....
```

Implementation source:
https://github.com/Kajabi/sage/blob/46af05bcb0bbcd27c61ed9dbbe2509b15c3c7520/lib/sage_rails/app/sage_components/sage_component.rb#L17-L29

AndrwM (changes requested):
Love how this is coming along! Left a few comments

app/views/examples/objects/page_heading/_preview.html.erb
AndrwM: Ah, okay I see – I think it would be best here to use a `content_for` snippet so that any html could be passed in, ping me and I'll show you how this works!

lib/sage-frontend/stylesheets/system/patterns/objects/_page_heading.scss
philschanely: @AndrwM I agree on color, but not quite on spacing. Here we could use `gap` on `__toolbar` to enforce the spacing of these elements in this context and thus avoid specifying spacing directly on the elements.

AndrwM: Great idea! 👍

lib/sage_rails/app/views/sage_components/_sage_page_heading.html.erb
philschanely: Lets def include the spacer option on the component! 🙌 

Future grid and spacing setup should come from parent/context but having that spacer as an option on the component is a win!

lib/sage_rails/app/views/sage_components/_sage_page_heading.html.erb
pixelflips: > This should not be a modifier; should it perhaps be `.sage-page-heading__lead`, or `__copy` or `__intro`?

Updated to __intro based on our conversation. Thanks again!

lib/sage_rails/app/views/sage_components/_sage_page_heading.html.erb
pixelflips: This is fantastic, will add.

app/views/examples/objects/page_heading/_preview.html.erb
pixelflips: @AndrwM thanks for the content_for intro. Definitely simplified and cleaned things up.

lib/sage-frontend/stylesheets/system/patterns/objects/_page_heading.scss
pixelflips: @philschanely now extended. ❤️

app/views/examples/objects/page_heading/_preview.html.erb
pixelflips: @philschanely these we're greatly reduced and no longer required. @AndrwM had some thoughts and will reach out.

lib/sage-frontend/stylesheets/system/patterns/elements/_breadcrumbs.scss
```diff
@@ -9,7 +9,7 @@
 .sage-breadcrumbs__icon {
   display: inline-flex;
   align-items: center;
-  margin-right: rem(4px);
+  margin-right: rem(8px);
```
philschanely: Could we use `gap` on the container instead of `margin-right` here on the items?

lib/sage-frontend/stylesheets/system/patterns/elements/_button.scss
```diff
@@ -84,6 +84,10 @@
     padding-bottom: rem(9px);
     border-radius: 0;
   }
+
+  .sage-page-heading__actions & {
+    margin: 0;
```
philschanely, line 89: 😎

lib/sage-frontend/stylesheets/system/patterns/objects/_page_heading.scss
```diff
@@ -5,18 +5,75 @@
...
 
 .sage-page-heading__title {
   @extend %t-sage-heading-1;
+  grid-area: title;
+}
 
-  order: 2;
+.sage-page-heading__intro {
+  grid-area: intro;
+  margin-top: sage-spacing(sm);
```
philschanely, line 26: Can `grid-gap` be used on the container instead of these `margin-x` settings?

lib/sage-frontend/stylesheets/system/patterns/objects/_page_heading.scss
```diff
@@ -5,18 +5,75 @@
...
+      @include sage-icon-base($-icon-name);
+    }
+  }
+}
+
+.sage-page-heading__actions {
+  display: flex;
+  grid-area: actions;
+  grid-gap: sage-spacing(sm);
+  align-items: flex-start;
```
philschanely, line 57: Could these settings that change responsively be wrapped in a media query that boxes them to the space they're used? This may help make it more explicit what is changing responsively versus what is constant. 

In this case you might have this:
```
.sage-page-heading__actions {
  grid-area: actions;
  display: flex;
  gap: sage-spacing(sm);
  // Hopefully margin can be removed in favor of a grid-gap setting on the header itself
  // margin-top: sage-spacing(sm);

  @media (max-width: sage-breakpoint(sm-max)) {
    flex-wrap: wrap;
    justify-content: flex-start;
  }

  @media (min-width: sage-breakpoint()) {
    flex-wrap: nowrap;
    justify-content: flex-end;
  }
}
```

lib/sage-frontend/stylesheets/system/patterns/objects/_page_heading.scss
```diff
@@ -5,18 +5,75 @@
 ================================================== */
 
 .sage-page-heading {
-  display: flex;
-  flex-flow: column;
-  margin-bottom: sage-spacing(lg);
+  margin-bottom: sage-spacing(sm);
+  @media (min-width: sage-breakpoint()) {
```
philschanely, line 8: Per my comment below about using grid-gap for margins, you might choose to revise to this here:

```
.sage-page-heading {
  display: grid;
  margin-bottom: sage-spacing(sm);

  @media (max-width: sage-breakpoint(sm-max)) {
    grid-template-columns: auto;
  }

  @media (min-width: sage-breakpoint()) {
    grid-template-areas:
      "crumbs ."
      "title actions"
      "intro ."
      "toolbar secondary";
    grid-template-columns: 1fr auto;
  }
}
```

philschanely (commented):
A few final suggestions here. Looking really good!

lib/sage-frontend/stylesheets/system/patterns/elements/_breadcrumbs.scss
pixelflips: @philschanely good call, will adjust.

lib/sage-frontend/stylesheets/system/patterns/objects/_page_heading.scss
pixelflips: @philschanely quick question. I initially tried using the gap on the container, but this would add a gap to all the tracks which caused the spacing/gap to remain even if the area had no content. Is there a workaround you might know of?

lib/sage-frontend/stylesheets/system/patterns/objects/_page_heading.scss
pixelflips: @philschanely spacing should now be handled in the component, I just forgot to remove the margin so thanks for pointing that out.

A question around the media queries. Currently, I don't have grid applied to smaller screens since the different sections stack up nicely without. Could you clarify as to why I might need to update that just so I better understand?

philschanely (approved):
Spot on, dude! :tada

PR #414, opened 2020-09-16
Products Navbars Bugfixes

AndrwM:
## Description
- SageDropdown: In Products Admin Navbars where dropdown were not anchored on the right
  - It looks like this may have been a misnamed class in a recent upgrade of SageDropdown but I could use a sanity check here
-  SageModals: In Products modals were not able to be closed with the cancel button
   - This is due to a recent refactor of SageButton to permit button text truncation 

----

The Sage local link script for "frontend-only" didn't have the right file permissions to run, this has also been fixed.


philschanely (approved):
Nice. Thanks for the fixes and sorry for my oversight.

PR #415, opened 2020-09-16
New panels and cards structure, Part 1

philschanely:
## Description

This PR is Part 1 of an update to better distinguish Panels from Cards. It accomplishes some preliminary cleanup as follows:

- Creates Expandable Card from old Accordion Panels variations.
- Creates Data Card from old Panels variations.
- Creates Empty State from old Card variations.
- Tidies what remains in Panels in preparation for next PR.
- Deprecates Card as it stands to make room for the new card setup in next PR.

Substantive changes must be made in `products` to accompany this work, much of which is drafted in separate branches there.


pixelflips (approved):
👍  +1 - Pulled this down and saw no issues that jumped out. LGTM! Nice work!

joe-dakroub (approved):
No concerns here. Nice work Phil. 👍

PR #416, opened 2020-09-17
SageCatalogCard (+ SageRails Docs support)

AndrwM:
# About 🧑🏻‍💻
**Preview:** https://sage-build-244-ajm-cata-tric9f.herokuapp.com/pages/breakout/object/catalog_card

This PR includes updates to Docs code preview strategy, allowing the use and development of SageRails components as a first-class citizen in the Docs. This is a move to _exclusively_ use the SageRails components in Docs when using a Sage component.

### Components Added
- SageCatalogCard
  - Used for display of lists of Offers and Products and likely more

### Additional:
- Support for choosing whether to us our "legacy" html-only previews or SageRails-based previews
- Display SageRails previews and html source originating from SageRails component
- Refinements to SageButton--tiny modifier
- Sage border light token
- Update sage label font-size to match mocks
- Allow multiple code previews to exist on the same page

# Screencap
image
image

# Comps
image
image


lib/sage-frontend/stylesheets/system/patterns/objects/_catalog_card.scss
```diff
@@ -0,0 +1,83 @@
...
+  type: object
+
+
+================================================== */
+
+$_content-height: 100px;
+$_image-height-mobile: 200px;
+
+.sage-catalog-card {
+  padding: sage-spacing(sm) sage-spacing(sm) sage-spacing(sm) sage-spacing(md);
```
philschanely: Double-check this with Sam as he may desire for this to be consistently 16px inside card. Then consider nudging the options element itself into the gutter if needed?

lib/sage-frontend/stylesheets/system/patterns/objects/_catalog_card.scss
```diff
@@ -0,0 +1,83 @@
...
+  background-color: sage-color(white);
+  border-bottom: sage-border();
+
+  display: grid;
```
philschanely, line 13: 🙌

lib/sage-frontend/stylesheets/system/patterns/objects/_catalog_card.scss
```diff
@@ -0,0 +1,83 @@
...
+  grid-template-columns: minmax(100px, 1fr) 5fr min-content;
+  grid-template-rows: repeat(2, ($_content-height / 2));
+  grid-column-gap: sage-spacing(md);
+  grid-row-gap: sage-spacing(xs);
+  grid-template-areas:
```
philschanely: Would you mind guarding the responsive props with a `min-width` for the aspects that change above sage-breakpoint? IMO this makes the overrides/changes clearer and simpler to trace properties in the inspector.

lib/sage-frontend/stylesheets/system/patterns/objects/_catalog_card.scss
```diff
@@ -0,0 +1,83 @@
...
+$_image-height-mobile: 200px;
+
+.sage-catalog-card {
+  padding: sage-spacing(sm) sage-spacing(sm) sage-spacing(sm) sage-spacing(md);
+  background-color: sage-color(white);
+  border-bottom: sage-border();
+
+  display: grid;
+  grid-template-columns: minmax(100px, 1fr) 5fr min-content;
+  grid-template-rows: repeat(2, ($_content-height / 2));
```
philschanely: Any reason not to let these be auto or omit altogether?

lib/sage-frontend/stylesheets/system/patterns/objects/_catalog_card.scss
```diff
@@ -0,0 +1,83 @@
...
+$_content-height: 100px;
+$_image-height-mobile: 200px;
+
+.sage-catalog-card {
+  padding: sage-spacing(sm) sage-spacing(sm) sage-spacing(sm) sage-spacing(md);
+  background-color: sage-color(white);
+  border-bottom: sage-border();
+
+  display: grid;
+  grid-template-columns: minmax(100px, 1fr) 5fr min-content;
```
philschanely: This is 👍  but I interpreted it as `auto 1fr auto`... will that suffice? Do we actually want the image to scale? If so consider instead defining an absolute upper boundary such as `minmax(100px, 200px) 1fr auto`. If not, consider just `120px 1fr auto` if that image does need a bounded size. Just trying to reduce any magic-ness and bind the possibilities if possible.

lib/sage-frontend/stylesheets/system/patterns/objects/_catalog_card.scss
```diff
@@ -0,0 +1,83 @@
+/* ==================================================
+  ** catalog-card
+  type: object
+
+
+================================================== */
+
+$_content-height: 100px;
```
philschanely: is the `$_` something new we should know about? We've been using `$-`

lib/sage-frontend/stylesheets/system/patterns/objects/_catalog_card.scss
```diff
@@ -0,0 +1,83 @@
...
+  grid-template-rows: repeat(2, ($_content-height / 2));
+  grid-column-gap: sage-spacing(md);
+  grid-row-gap: sage-spacing(xs);
+  grid-template-areas:
+  "img body actions"
+  "img meta actions";
+}
+
+@media (max-width: sage-breakpoint(sm-min)) {
+  .sage-catalog-card {
```
philschanely: 😍   responsive grid!!

lib/sage-frontend/stylesheets/system/patterns/objects/_catalog_card.scss
```diff
@@ -0,0 +1,83 @@
...
+}
+
+.sage-catalog-card__img {
+  grid-area: img;
+  overflow: hidden;
+  background: sage-color(grey, 200);
+  position: relative;
+  border-radius: sage-border(radius);
+
+  img {
```
philschanely, line 67: Consider instead:

```
width: 100%;
height: 100%;
object-fit: cover;
```

app/views/pages/sandbox.html.erb
```diff
@@ -2,7 +2,43 @@
...
+              icon: { name: "dot-menu-horizontal", style: "only" },
+              value: "Drop",
+              align: "end"
+            } %>
+          <% end %>
+        </div>
+        <div class="sage-catalog-card__body">
+          <a class="t-sage-heading-5" tabIndex="0" href="#">This is a SageCatalogCard</a>
+        </div>
+        <div class="sage-catalog-card__meta">
```
philschanely: Could this be a `p` or allow type safely inside?

app/views/pages/sandbox.html.erb
```diff
@@ -2,7 +2,43 @@
...
+          } do %>
+            <%= sage_component SageButton, {
+              style: "tertiary",
+              icon: { name: "dot-menu-horizontal", style: "only" },
+              value: "Drop",
+              align: "end"
+            } %>
+          <% end %>
+        </div>
+        <div class="sage-catalog-card__body">
```
philschanely: Could this be a `p` or allow type safely inside?

app/views/pages/sandbox.html.erb
```diff
@@ -2,7 +2,43 @@
...
-<%= sage_component SagePanel, {} do %>
-  <%= sage_component SageButton, { style: "primary", value: "This is a sage component rails test" } %>
-<% end %>
+
+
+      <section class="sage-catalog-card">
+        <div class="sage-catalog-card__img">
+          <img alt="Alternative text for image" src="/packs/media/images/docs/card/card-placeholder-sm-793ef84766b339810b660be77cb81fdd.png">
+        </div>
+        <div class="sage-catalog-card__actions">
```
philschanely: Do we want the actions here in the source order? I would think the screen reader would prefer the content itself first. I imagine the image could even be last in the source order. 🎉  Grid areas!!

lib/sage-frontend/stylesheets/system/patterns/objects/_catalog_card.scss
```diff
@@ -0,0 +1,84 @@
+/* ==================================================
+  ** catalog-card
+  type: object
+
+
+================================================== */
+
+$_image-height: 64px;
```
philschanely: Is the `$_` something new we should use? We have `$-` elsewhere. And be sure any of your `px` values are wrapped in `rem()`

lib/sage-frontend/stylesheets/system/patterns/objects/_catalog_card.scss
```diff
@@ -0,0 +1,84 @@
...
+  grid-row-gap: sage-spacing(xs);
+  grid-template-areas:
+  "img body actions"
+  "img meta actions";
+}
+
+@media (max-width: sage-breakpoint(sm-min)) {
+  .sage-catalog-card {
+    grid-template-rows: $_image-height-mobile auto auto;
+    grid-template-areas:
```
philschanely, line 26: You could probably switch to two columns here and update the configs accordingly.

lib/sage-frontend/stylesheets/system/patterns/objects/_catalog_card.scss
```diff
@@ -0,0 +1,84 @@
...
+  align-items: flex-end;
+  overflow: hidden;
+
+  > * {
+    @include sage-truncate;
+  }
+
+}
+
+@media (max-width: sage-breakpoint(sm-min)) {
```
philschanely: Lets nest this inside `.sage-catalog-card__body` and then you don't need to repeat the selector.

lib/sage-frontend/stylesheets/system/patterns/objects/_catalog_card.scss
```diff
@@ -0,0 +1,84 @@
...
+  grid-template-columns: minmax(100px, $_image-width) auto min-content;
+  grid-template-rows: repeat(2, ($_image-height / 2));
+  grid-column-gap: sage-spacing(md);
+  grid-row-gap: sage-spacing(xs);
+  grid-template-areas:
+  "img body actions"
+  "img meta actions";
+}
+
+@media (max-width: sage-breakpoint(sm-min)) {
```
philschanely: Lets nest this inside `.sage-catalog-card` and then you don't need to repeat the selector.

lib/sage-frontend/stylesheets/system/patterns/objects/_panel.scss
```diff
@@ -103,6 +103,11 @@ $-panel-spacing-sm: sage-spacing(xs);
   .sage-panel--flat-sm > & {
     padding: $-panel-spacing-sm;
   }
+
+  .sage-panel--content-flush > &,
```
philschanely: Consider instead letting the contents use negative margin to such them into the gaps? Regardless, may need `overflow: hidden` for rounded corners to show.

philschanely (changes requested):
Stellar work with the grid stuff, @AndrwM 😍 ! Some changes suggested below. Happy to discuss as needed.

lib/sage-frontend/stylesheets/system/patterns/objects/_catalog_card.scss
AndrwM: Whoops, nope!

lib/sage-frontend/stylesheets/system/patterns/elements/_info.scss
```diff
@@ -0,0 +1,27 @@
...
+
+
+================================================== */
+
+.sage-info {
+  @extend %t-sage-body-small;
+
+  color: sage-color(charcoal, 200);
+  display: inline-flex;
+  margin: 0 sage-spacing(sm) 0 0;
```
philschanely: Instead of this default margin, could we instead add it in the context of being inside a certain container? For example, you could here do `.sage-catalog-card__info & { margin-right: sage-spacing(sm); }`. Or perhaps consider:

```
.sage-info-group {
  display: inline-flex;
  gap: sage-spacing(sm);
}
```
You could also consider just putting this rule on `.sage-catalog-card__info` itself too and avoid the extra container as a result.

Bottom line: this is a great example of where we can resist the urge to put on default margin and instead use context to add it when needed.

lib/sage-frontend/stylesheets/system/patterns/elements/_info.scss
```diff
@@ -0,0 +1,27 @@
...
+}
+
+@each $icon-name, $icon-code in $sage-icons {
+  .sage-info--icon-#{$icon-name} {
+
+    &::before {
+      @include icon-base($icon-name, md);
+
+      align-self: center;
+      margin: 0 sage-spacing(xs) 0 0;
```
philschanely: Is it necessary to specify all sides at `0` or would just `margin-right` only suffice?

lib/sage-frontend/stylesheets/system/patterns/objects/_catalog_card.scss
```diff
@@ -0,0 +1,79 @@
...
+  background-color: sage-color(white);
+  display: grid;
+  grid-column-gap: sage-spacing(md);
+  grid-row-gap: sage-spacing(xs);
+
+  &:not(:last-child) {
+    border-bottom: sage-border(light);
+  }
+
+  @media (min-width: sage-breakpoint()) {
```
philschanely: 🙌 And this should probably switch to `@media (min-width: sage-breakpoint(md-min)) { ...`

lib/sage-frontend/stylesheets/system/patterns/objects/_catalog_card.scss
AndrwM: This is causing the image to expand the column, reverted back to using the center position absolute solution. In this situation its more important to show the images in a uniform size than to show the entirety of the image.

lib/sage-frontend/stylesheets/system/patterns/objects/_catalog_card.scss
AndrwM: Just for clarity on which breakpoint its using? Right on, will do

lib/sage-frontend/stylesheets/system/patterns/elements/_button.scss
```diff
@@ -108,20 +108,21 @@
 }
 
 .sage-btn--small {
-  padding-top: 0;
-  padding-bottom: 0;
+  padding: 0 sage-spacing(sm);
```
philschanely, line 111: May not be necessary to override left and right spacing here. Since this is a modifier it only needs to tap the settings that are changing.

lib/sage-frontend/stylesheets/system/patterns/objects/_catalog_card.scss
```diff
@@ -0,0 +1,78 @@
...
+================================================== */
+
+$-image-height-min: rem(64px);
+$-image-width: rem(120px);
+$-image-height-mobile: rem(120px);
+
+.sage-catalog-card {
+  display: grid;
+  grid-column-gap: sage-spacing(md);
+  grid-row-gap: sage-spacing(2xs);
```
philschanely: Design specs seem to me to show 8px gap rather than 4px. Try that instead and then use `align-self: end` and `align-self: start` to `title` and `content` respectively. I wish there was a `pack-center` option but this may have to do for now.

lib/sage-frontend/stylesheets/system/patterns/objects/_catalog_card.scss
philschanely: Hmm I see. 🤔  Pros and cons on both sides. Seems this is also possible with a `max-height` on `__image` and then flex to center the image in the space. I suppose it comes down to why one or the other and I don't have a good reason other than trying to avoid absolute positioning when possible. This is certainly a feasible use for abs pos though, and I think that's better than adding a height dependency unless you think that benefits the situation.

philschanely (approved):
Few small suggestions below. Excellent work on this stuff.

lib/sage-frontend/stylesheets/system/patterns/objects/_catalog_card.scss
AndrwM: I'd prefer to not use `max-height` with this to allow for the image container (and image) to grow to allow multiple lines of `sage-catalog-card__content`.

Example, in a smaller viewport with a lot of items:
image



I generally agree with you on absolute position though this approach seems to give us what we need. I'm down to pair on this, maybe I'm missing something here 😄

PR #418, opened 2020-09-22
Docker local development

voodooGQ:
## Description
This PR adds Docker Support to the Sage repository for local development. More information can be found in the `docker/README.md` file on setting up.

### Steps for testing
1. Download Repo to existing computer
1. Follow all Prereqs in the `docker/README.md` document
1. Profit $$$

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -1,0 +1,36 @@
+# Docker Local Development
+
+The repo includes a `docker-compose.yml` config for running the app stack
+locally with Docker. This is useful if you want to get the app running quickly.
+
+## Pre-requisites
+
+1. Create an ssh-key (`ssh-keygen`)
+1. [Add your SSH key to Github](https://docs.github.com/en/github/authenticating-to-github/adding-a-new-ssh-key-to-your-github-account)
+1. Clone this repo
+1. Add entries to your host file by running `sudo docker/scripts/hosts.sh`, you will be prompted for your administrator password.
+
+### Mac
+
+1. [Homebrew](https://brew.sh/)
+1. Git (`brew install git`)
+1. [Docker for Mac](https://docs.docker.com/docker-for-mac/install/)
+
+### Linux (Ubuntu)
+
+1. `sudo apt-get install git docker.io docker-compose`
+1. `sudo service docker start`
+1. `sudo chmod 666 /var/run/docker.sock`
+
+## Spinning Up
+
+1. Boot the app with `docker-compose up` (☕️ this will take a while the first time while dependencies install)
+1. Hit http://www.sage.test when booted
+
+Docker will bind to port 80, and 4000, so those ports need to be available on your
+machine.
+
+Dependencies and migrations are kept up to date automatically every time the
+containers are started. If you do need a command prompt to run tasks,
+use `docker-compose exec web bash` while the containers are running to connect
+to the running web container.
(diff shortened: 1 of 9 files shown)


docker/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+### Mac
+
+1. [Homebrew](https://brew.sh/)
+1. Git (`brew install git`)
+1. [Docker for Mac](https://docs.docker.com/docker-for-mac/install/)
+
+### Linux (Ubuntu)
+
+1. `sudo apt-get install git docker.io docker-compose`
```
philschanely, line 21: 🤓

docker/scripts/hosts.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+# Adds entries to the host file for sage.test and
+# kajabi-sage.test if missing
```
philschanely, line 3: Just curious what the difference or use would be for `sage.test` versus `kajabi-sage.test`?

philschanely (approved):
Ive not tested getting this set up locally yet, but looks good otherwise!

docker/scripts/hosts.sh
voodooGQ: They will point to the same thing in the end. We're using `kajabi.test` for the main products repo so I wanted to add a host entry that reflects a similar naming scheme. If you'd prefer we can remove it and just keep `sage.test`

teenwolfblitzer:
@voodooGQ I'm running ruby-2.4.9 on sage using chruby but have the following error when trying to spin up. Any ideas?

```
web_1     | Install missing gems with `bundle install`
web_1     | Fetching gem metadata from https://rubygems.org/.............
web_1     | byebug-11.1.3 requires ruby version >= 2.4.0, which is incompatible with the
web_1     | current version, ruby 2.3.8p459
web_1     | yarn run v1.13.0
web_1     | $ npm-run-all --parallel webpack:server rails:server
web_1     | $ ./bin/webpack-dev-server
web_1     | $ rails server -p 4000 -b 0.0.0.0
web_1     | /bin/sh: 1: rails: not found
web_1     | error Command failed with exit code 127.
web_1     | info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
web_1     | ERROR: "rails:server" exited with 1.
web_1     | error Command failed with exit code 1.
web_1     | info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

voodooGQ:
@monkeypox8 Interesting, I'm wondering why it's using Ruby version 2.3.8 on the container, it should only have 2.4.9 installed.

Could you do me a favor and run the following: `docker-compose up >> docker.log` and attach the `docker.log` file here? Thank you!

teenwolfblitzer:
> @monkeypox8 Interesting, I'm wondering why it's using Ruby version 2.3.8 on the container, it should only have 2.4.9 installed.
> 
> Could you do me a favor and run the following: `docker-compose up >> docker.log` and attach the `docker.log` file here? Thank you!

Sure thing, here's the log. The script was hanging after `sage_web_1` so I exited out after about 5 min. It does seem to be referencing ruby 2.3.8, though if I run `which ruby`, the resulting output is 2.4.9.

[docker.log](https://github.com/Kajabi/sage/files/5270554/docker.log)

joe-dakroub (approved):
I tested this locally and it works very well. Nice work and thank you! 💯 👍

teenwolfblitzer (approved):
👍 🚢  Thanks for getting me back up and running on this, @voodooGQ, working great!

PR #420, opened 2020-09-22
SageLabel Rails, update Docs preview for SageDropdown, minor adjustments

AndrwM:
### New
- SageRails: SageLabel component

### Updates
- Rename `CatalogCard` -> `CatalogItem`
- SageLabels: Update to be 14pt type instead of 12pt (per new design updates)
- SageButtons: Prevent from collapsing in height if an empty string is provided
- SageDropdown: Rename `item[:name]` to `item[:value]` to match implementation of other components that accept item arrays.

### Docs
- SageDropdown: component preview conversion


app/views/examples/objects/catalog_item/_preview.html.erb
```diff
@@ -23,11 +23,35 @@
       value: "1 Partridge In A Pear Tree",
     } %>
     <%= content_for :sage_aside do %>
-      <span class="sage-label sage-label--draft">Draft</span>
       <%= sage_component SageDropdown, {
         align: "right",
         items: [{
-          name: "Edit Item",
+          value: sage_component(SageLabel, { color: "success", value: "Publish" }),
```
philschanely, line 29: What would you think of calling this `label`, `text`, or `content` instead of `value`?

app/views/examples/objects/dropdown/_preview.html.erb
```diff
@@ -1,96 +1,114 @@
...
+   }, {
+    value: "Take A Dangerous Action",
+    icon: "remove-circle",
+    style: "danger",
+    attributes: { "href": "#" },
+  }]
+} do %>
+  <%= sage_component SageButton, {
+    style: "primary",
+    value: "Add An Element"
```
philschanely, line 26: Another place where `value` doesn't feel like the right key for the content to me. I guess I tend to see this as `label` or `text`. 🤔

pixelflips (approved):
LGTM! Love seeing sage_component being used! 👍

lib/sage_rails/app/views/sage_components/_sage_catalog_item.html.erb
```diff
@@ -1,21 +1,21 @@
...
   <% end %>
   <% if content_for? :sage_aside %>
-    <aside class="sage-catalog-card__aside">
+    <aside class="sage-catalog-item__aside">
       <%= content_for :sage_aside %>
     </aside>
   <% end %>
   <%# Uses tabindex="-1" because there are two links to the same location in this component (component.title also links to the href) %>
-  <a class="sage-catalog-card__img" href="<%= component.href %>" tabindex="-1" title="Go to: <%= component.title %>">
+  <a class="sage-catalog-item__img" href="<%= component.href %>" tabindex="-1" title="Go to: <%= component.title %>">
```
philschanely, line 18: Thinking out loud, will the image always be hyperlinked, or is there a situation where we might want just a `div` wrapper here instead?

philschanely (commented):
A few suggestions below but nothing blocking per se. 🙌

philschanely:
@AndrwM  I Should have said too, "Thank you!" for doing the name updates. Also seconding @pixelflips that its great to see the Rails components being worked in.

lib/sage_rails/app/views/sage_components/_sage_catalog_item.html.erb
AndrwM: As far as I know it will always be hyperlinked. 🤷‍♂️

app/views/examples/objects/dropdown/_preview.html.erb
AndrwM: Value is the current pattern we're using for a few components now. I think I prefer this over something like text because the naming allows for implementations like this:

```
<%= sage_component SageButton, {
    style: "tertiary",
    style: "tiny",
    icon: { style: "left", name: "copy" },
    value: sage_component(SageLabel, { color: "draft", value: "asdf" })
} %>
```

app/views/examples/objects/catalog_item/_preview.html.erb
AndrwM: For `SageLabel` I'm open to changing `value` to `content` or something else... This naming move was made for consistency. 

Currently SatusIcon, Label, Button, Dropdown use `value` as the argument for the "content".

AndrwM:
@philschanely and I resolved the above discussion in slack. All good

PR #421, opened 2020-09-23
New panels and cards structure, Part 3

philschanely:
## Description

This is the third and final part of work separating out Panels and Cards that adds the full styling and documentation around them. It includes the following:

- Panels and Cards page under Foundations that explains the difference
- Panel Object that shows panels and all the elements available within them
- Card Object that shows cards and all the elements available within them
- Two possible options for a Grid Template utility that can be placed on "row" elements such as `sage-card__row` and `sage-panel__row`. See Grid Template A and Grid Template B under Foundations. 
- Sandbox page shows Panel and Card with grid templates all in use.
- A few miscellaneous cleanup tasks

Tasks remaining:

- [x] Articulate method for providing custom grid templates
- [x] Update the Rails components to match these specs

**NOTE:** Sandbox and extra grid templates should be removed and code updated as a result before merging.


philschanely:
@Kajabi/ux-devs This is the PR for the panel vs. card work I started to demonstrate in our recent sync. Of note are the new elements available within Panels and Cards and the proposed Grid Templates. Please let me know what you think about those, and if either one seems more approachable than the other.

AndrwM:
@philschanely This is really strong work! I love the grid template idea. I think it will be really great to have this in here.

A few questions:
1. I'm not sure I'm understanding the purpose of `sage-card__copy`, is that an unstyled div that is a container for the full-width content?
2. `sage-card` seems to be mixing two concerns that may be better split:
    - padding and a visual border
    - The vertical alignment of elements within an area (a y-grid, almost)
3. Aside from `tiles` that exist in SagePanel but not SageCard, is there a difference between these two implementations?
4. Grid names, what if we took a more visually representational approach like this: `sage-grid-template-xxmmx`
    - `x` visually representing static
    - `m` with its longer letterform, visually representing fluid/auto
    - Another option could be `sage-grid-template-ssffs` (static/fluid)... but that doesn't read as well.
5. Within a SageGrid how will widths of elements be defined? (For example if I wanted an image to be a specific width on non-mobile viewports.) Would our approach approach be to either 1) created a custom component or 2) add a min-width directive as 1-off css in Products?
6. How backwards compatible is this with our current Kajabi-Products Sage code? What does an upgrade look like?

joe-dakroub (approved):
The documentation around Card and Panel is helpful and clear.

The grid templates are a great concept. I don't mind the morse code approach (though I don't know morse code) and think that in time, the names will become familiar enough that it will work well. 

I really like the dot and dash notation and wonder if that could be used in an easy way to specify layouts, something like `data-grid-template=".---"`? It is easy to visualize which template you are referring to. Just a thought.

Lots of thoughtful work here Phil! Thank you. 👍

philschanely:
Thanks, @Joe! Your idea of using `data-` is a great idea that might be an in-between for what @AndrwM was looking for too. I'll look into the implications.

AndrwM:
@joedakroub Oh, great idea!

philschanely:
@joedakroub and @AndrwM I have added the data-sage-grid-template option to include the dot and dash visual syntax but not applied it in all places where grid template is allowed. I think it needs further consideration but given this setup should be simple to refactor if we settled on that direction. Merging!

PR #424, opened 2020-09-25
SageOutline Item

AndrwM:
image


joe-dakroub (approved):
This is really nice work @AndrwM! Outline views can be difficult to pull off well, but this code is tight and really clear.

One thought to improve this for implementation (probably out of scope) would be to add the ARIA patterns for [treegrid](https://www.w3.org/TR/wai-aria-practices-1.1/examples/treegrid/treegrid-1.html).

Roughly, the ARIA attributes to use would be:
- Set `role="treegrid"` on the root element
- For each row:
  - Set `role="row"`
  - Set `aria-expanded=true/false` based on its expand/collapse state
  - Set `aria-level=#` based on the indentation level of the row
  - Set `aria-posinset=#` based on the position of the row in the set of indented items
  - Set `aria-setsize=#` based on the number of rows in the set of indented items
- For each row cell:
  - Set `role="gridcell"`

It is a somewhat involved process to set this data up, manage the state and add keyboard access, but I have a working example I can dig up that shows the complete pattern in use if/when this is something we would want to do in the future.

Super cool to see this in Sage!

AndrwM:
@joedakroub Oh, okay very cool! Thanks for the notes.

The final html implementation of this feature will be in React, I'll add those attributes within the JSX and dynamically calculate the nest depths. 🥳

AndrwM:
@joedakroub I'll reach out with the JSX stuff for you to take a look at once its complete!

joe-dakroub:
Thanks @AndrwM! I look forward to seeing the JSX work. 💯

lib/sage-frontend/stylesheets/system/patterns/objects/_outline_item.scss
```diff
@@ -0,0 +1,168 @@
...
+  type: object
+
+
+================================================== */
+
+$-collapse-breakpoint-key: md-max;
+
+.sage-outline-item {
+  overflow: hidden;
+  margin-bottom: sage-spacing(md);
```
philschanely: This margin should not be needed in the implementation inside of Panel since it uses `gap` to provide spacing between items. Bummer that for our docs site the `sage-type` container is there that interrupts the normal spacing. I hope to work on improving this structure soon for the docs!

However, I see an `8px` gap in the Products Admin mockup in Figma. Is the `24px` indeed desired instead? Otherwise, consider a wrapper for the set of outline items such as `sage-outline` or `sage-outline-list` and use `gap` in lieu of margin on the individual items.

lib/sage-frontend/stylesheets/system/patterns/objects/_outline_item.scss
```diff
@@ -0,0 +1,168 @@
...
+
+.sage-outline-item__content {
+  display: grid;
+  align-items: center;
+  grid-template-rows: repeat(auto-fit, auto);
+  grid-column-gap: sage-spacing(sm);
+  padding-right: sage-spacing(sm);
+  padding-left: sage-spacing(md);
+
+  @media (min-width: sage-breakpoint($-collapse-breakpoint-key)) {
```
philschanely, line 32: 😍 Nice work on these responsive breakpoints!

lib/sage-frontend/stylesheets/system/patterns/objects/_outline_item.scss
```diff
@@ -0,0 +1,168 @@
...
+    content: sage-icon(arrow-corner);
+    color: sage-color(primary);
+  }
+  /* stylelint-enable selector-max-compound-selectors */
+
+  /* stylelint-disable declaration-no-important */
+  // Universal Hover For All
+  .sage-outline-item__content:hover > & {
+    &::before {
+      content: sage-icon(handle) !important;
```
philschanely, line 145: I agree this is best for now! 🙌 No need to overthink it haha

philschanely (approved):
Echoing Joe, this is really great, elegant work, @AndrwM ! 🙇‍♂️ Well done! 🏆 

Just a comment about spacing other than praise below ;)

lib/sage-frontend/stylesheets/system/patterns/objects/_outline_item.scss
AndrwM: 👅 I kept wrestling with the specificity but due to the `--category` modifier this wound up being the most focused solution. Thanks for understanding!

lib/sage-frontend/stylesheets/system/patterns/objects/_outline_item.scss
AndrwM: Thanks @philschanely! I'll update. I agree, `gap` would be better here.

PR #426, opened 2020-09-29
SageSortable: Refactor, add "Card" and (implicit) "List" types

AndrwM:
# Description
SageSortable is currently only used in Navbars. This refactor updates SageSortable to have "large" and "standard" style variants.

## Usage

SageSortable Standard
- Used in Product Edit Downloads
- Used in Product Edit Attachments

Comps:
image

SageSortable Large
- Used in Navbars

Production Screenshot:
image

------------
------------
------------

## Preview Of New Changes
screencast 2020-09-29 11-55-00


lib/sage_rails/app/sage_components/sage_sortable_item.rb
```diff
@@ -1,6 +1,8 @@
 class SageSortableItem < SageComponent
   attr_accessor :url_update
-  attr_accessor :name
+  attr_accessor :title
+  attr_accessor :subtitle
   attr_accessor :label
   attr_accessor :id
+  attr_accessor :type
```
philschanely: I think `type` works... however, I'm not sold on implicit "regular" and then "large". What I see in the renderings looks more like a "list item" context and a "card" context based on the styles applied. Could you see revising to match this system styling context in some way?

teenwolfblitzer (commented):
This is looking great! Spotted one item to check out—since Safari doesn't yet support `gap` with flexbox, the title and subtitle run together on the normal variant:

flexbox-gap

philschanely (commented):
Overall looking great! Just a question about how we're labeling the two variations below.

lib/sage_rails/app/sage_components/sage_sortable_item.rb
AndrwM: That makes way more sense. I'll revise!

AndrwM:
@monkeypox8 Ooof, dang. Okay I'll look into a fallback. TY!

PR #427, opened 2020-09-29
Tooltip Sizing Adjustment 

pixelflips:
## Description
When having a longer amount of text within a tooltip the text currently does not wrap as expected due to the size being affected by the `min-width: max-content`. 

This update removes the `min-width`, which resolves the issue and causes the text to wrap based on the default width applied or with the option `data-js-tooltip-size`.

### Screenshot
|  before  |  after  |
|--------|--------|
|Screen Shot 2020-09-29 at 1 31 02 PM|Screen Shot 2020-09-29 at 1 31 25 PM|

## Test notes
You will have to manually add a long string of text to the `/tooltip/_preview.html.erb` file or you may also update the text within the `data-js-tooltip` applied to any of the tooltip elements using dev tools. 

## Related
- NA

### DIFF
--- a/lib/sage-frontend/stylesheets/system/patterns/elements/_tooltip.scss
+++ b/lib/sage-frontend/stylesheets/system/patterns/elements/_tooltip.scss
@@ -8,7 +8,6 @@
   display: flex;
   position: absolute;
   z-index: $sage-tooltip-zindex;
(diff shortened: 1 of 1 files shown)


AndrwM (approved):
Texas-sized +1

👍

PR #431, opened 2020-10-01
Add button group and revise to use rails component

philschanely:
## Description

https://kajabi.atlassian.net/browse/MAN-634?atlOrigin=eyJpIjoiODYwMTk4YTRmYWUxNDNhMWE5MGFmY2FkN2M1YmFjZTciLCJwIjoiaiJ9

Adds a `sage-btn-group` class and SageButtonGroup rails component for applying spacing between buttons and adjust alignment when needed.

See revised Elements > Buttons page.

NOTE: Changes necessary in Products drafted here https://github.com/Kajabi/kajabi-products/pull/15872

ALSO: Apologies; I accidentally pushed this work directly to `master` but then reverted to do the proper thing here. 😊


joe-dakroub (approved):
Looking good. No concerns. Nice workaround on the `gap` issue for Safari. 👍

PR #434, opened 2020-10-01
Color palette and placeholder text color updates

pixelflips:
## Description
Updates colors in the color palette based on a request from design. 
In addition, updates the placeholder text color for inputs and textarea for better contrast between placeholder and values. 

**Updated colors:**
- yellow/400 = #8d5c2f
- yellow/500 = #5c381e
- charcoal/100 = #65778b
- charcoal/200 = #526275
- placeholder text = charcoal/100

## Related
- NA


teenwolfblitzer (commented):
Looking great! I noticed one minor item with disabled inputs. It's more visible with the input textarea example since it uses a predefined value. Disabled inputs with just a placeholder appear as expected:

input-disabled

But on a disabled input with a pre-filled value, the disabled text color (set in `_variables.scss`) remains the lighter grey shade:
textarea-disabled

pixelflips:
@monkeypox8 - a great catch 🎣 . will adjust the disabled input to match the textarea.

teenwolfblitzer (approved):
👍 🔥 LGTM, thanks for jumping on this one!

PR #437, opened 2020-10-05
Add copy button, text, and card

philschanely:
## Description

Adds new styles for SageCopyText (inline), SageCopyTextCard (block) and an actual SageCopyButton (`sage-copy-btn`) to handly succinct styling for these related visual elements.

See Elements > Copy Text

### Screenshots

Screen Shot 2020-10-05 at 12 51 53 PM

Non-blocking related PR drafted here: TBD

### DIFF
--- a/app/helpers/elements_helper.rb
+++ b/app/helpers/elements_helper.rb
@@ -12,14 +12,6 @@ module ElementsHelper
   def sage_elements
     [
       # Sage Generated Elements
-        {
-          title: "copy_text",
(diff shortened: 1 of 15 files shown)


pixelflips (approved):
👍  - this is great. No concerns here and thanks so much for taking this on. I have a project coming up that will make use of these! Great work!

pixelflips:
@philschanely one minor thing I noticed is that the copy text (block) when using a `p` element is set to bold. Not a major issue, but in an upcoming update I'll be working on the text has a combo of bold and non-bold. I could use a different element for the text, such as a `div`, but just wanted to point that out and get your thoughts... not a show stopper by any means!

pixelflips:
@philschanely actually on closer look the `font-weight: 600` is set in the `sage-copy-text-card` styles. That will make it a bit tough to have a combo of bold and regular font styles.

philschanely:
@pixelflips I updated the weight settings so we should have a little more flexibility now. Also updated line spacing.

PR #439, opened 2020-10-06
Update SageRails SagePagination component

alimi:
# Description

In https://github.com/Kajabi/kajabi-products/pull/15773, we're changing the `pager` helper so it no longer requires nor accepts `pager_params`. It wasn't needed in most cases, and Kaminari deprecated the way `pager_params` was being used. Instead, `pager` will optionally accept `additional_params`.

This PR updates `SagePagination` so it also optionally accepts `addtional_params`. It'll match `pager`'s behavior in kajabi-products and clear up the Kaminari deprecation warning.`additional_params` defaults to an empty hash so it plays nicely with Kaminari.

`SagePagination` still exposes the ability to set `pager_params` via `#pager_params=`. When `#pager_params=` is called, `additional_params` will be set instead. This will make it easier to switch from `pager_params` to `additional_params` in kajabi-products because both the old and new API will be supported.

## Test notes

There shouldn't be any change in behavior in any pagination links.

## Related
- https://github.com/Kajabi/kajabi-products/pull/15773

### DIFF
--- a/lib/sage_rails/app/sage_components/sage_pagination.rb
+++ b/lib/sage_rails/app/sage_components/sage_pagination.rb
@@ -1,7 +1,16 @@
 class SagePagination < SageComponent
   attr_accessor :items
   attr_accessor :window
-  attr_accessor :pager_params
+  attr_accessor :additional_params
+
+  def initialize(attributes = {})
+    super
+    self.additional_params ||= {}
+  end
+
+  def pager_params=(new_pager_params)
+    self.additional_params = new_pager_params
+  end
 
   def prev_text
     "Back".html_safe
--- a/lib/sage_rails/app/sage_components/sage_pagination.rb
+++ b/lib/sage_rails/app/sage_components/sage_pagination.rb
@@ -8,10 +8,6 @@ class SagePagination < SageComponent
     self.additional_params ||= {}
   end
 
-  def pager_params=(new_pager_params)
-    self.additional_params = new_pager_params
-  end
-
   def prev_text
     "Back".html_safe
   end
--- a/lib/sage_rails/app/views/sage_components/_sage_pagination.html.erb
+++ b/lib/sage_rails/app/views/sage_components/_sage_pagination.html.erb
@@ -12,7 +12,7 @@
       </li>
       <% else %>
         <li class="sage-pagination__item">
-          <%= link_to_previous_page component.items, component.prev_text, class: "sage-pagination__page", params: component.pager_params %>
+          <%= link_to_previous_page component.items, component.prev_text, class: "sage-pagination__page", params: component.additional_params %>
         </li>
       <% end -%>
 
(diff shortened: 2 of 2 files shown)


lib/sage_rails/app/sage_components/sage_pagination.rb
```diff
@@ -1,7 +1,16 @@
...
+  attr_accessor :additional_params
+
+  def initialize(attributes = {})
+    super
+    self.additional_params ||= {}
+  end
+
+  def pager_params=(new_pager_params)
+    self.additional_params = new_pager_params
+  end
```
AndrwM: Is this mapping `pager_params` to `additional_params`?

We would prefer to update Kajabi-Products' implementation on Sage version-bump rather than maintaining backward compatibility with the `pager_params`.

AndrwM (changes requested):
We would prefer to handle the transition in Kajabi-Products than maintain backward compatibility. Would you mind updating?

Overall, the UXD team is excited to see PRs from outside our team being made to this repo. Thanks for digging into this. ✋

lib/sage_rails/app/sage_components/sage_pagination.rb
philschanely: Agreed, @AndrwM. Yet, in a case like this as a very kind effort from @alimi to make an improvement, would we let this go through as-is but make our own note to follow up with the correction and requisite fix on `products`? Otherwise its unclear to me who will take responsibility to patch on `products`. 🤔

teenwolfblitzer: If I'm understanding correctly, the process would be similar in either approach. Either way, we'd need to update the old `pager_params` references. The difference being how soon we'd want to have that done, since the current implementation may maintain compatibility, but also presents the opportunity for those unfamiliar with the change (more of a risk post-Sagification) to keep reusing the same pattern.

alimi: > We would prefer to update Kajabi-Products' implementation on Sage version-bump rather than maintaining backward compatibility with the `pager_params`.

Ah, ok. I added the backwards compatibility so kajabi-products would continue working whenever Sage was upgraded. But if we're gonna update kajabi-products whenever Sage is upgraded, that isn't really a concern. I can remove it.

> Otherwise its unclear to me who will take responsibility to patch on `products`. 🤔

I guess this would fall on whoever does the next Sage upgrade in kajabi-products? I'm still learning things and I'm not sure how the upgrade process works.

Also, [`SagePagination` is lightly used](https://github.com/Kajabi/kajabi-products/search?q=SagePagination) so it shouldn't be to hard to update when the time comes.

AndrwM (approved):
@alimi +1 Thanks for updating this! I'll plan to include this in v1.15.1 (targeting EOW)

jacobthemyth (approved):

PR #441, opened 2020-10-07
Misc rails comp tweaks

philschanely:
## Description

Makes a series of small tweaks as follows:

- Allow for hyperlinks in tab and choice
- Adjust width and truncation on Copy Text Card
- Add breadcrumbs as sage component
- Conform SageBody to `css_classes` prop. **NOTE:** Causes breaking change on `products` until the implementation of this component is synched with this change.
- Update panel divider full bleed and add card divider


pixelflips (approved):
👍  LGTM! Thanks for those breadcrumb updates! I think @joedakroub and I will be able to use those right away! ❤️  Great work!

PR #442, opened 2020-10-07
Revise SageOutline to allow all elements to live at the same html depth

AndrwM:
This reduces style logic that is better handled with react logic...
- Revise SageOutline to allow all elements to live at the same html depth
- Explicitly take an icon arg rather than attempting to add the correct "folder" icon depending on depth
- Remove the "Card" concept visually. This can be inserted into a card now.
- Add additional states that handle dragging and droptarget visual treatments
- Force outline items to grow to this size of their parent container (required for how React is calculating drag and drop locations)

Adds 6 new modifiers:
`sage-outline--category`
`sage-outline--depth-<number>`
`sage-outline--is-dragging`
`sage-outline--is-dragging-descendant`
`sage-outline--is-droptarget`
`sage-outline--is-droptarget-invalid`

Screencap:
image

### DIFF
--- a/app/views/examples/objects/outline_item/_preview.html.erb
+++ b/app/views/examples/objects/outline_item/_preview.html.erb
@@ -1,6 +1,9 @@
 <% 3.times do %>
   <%= sage_component SageOutlineItem, {
     title: "Wonderful Denim Pants",
+    icon: "folder",
+    depth: 0,
+    category: true,
     status: sage_component(SageLabel, { value: "Draft", color: "draft" }),
     actions_primary: [sage_component(SageButton, {
       value: "Add Content",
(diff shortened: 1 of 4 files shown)


philschanely (commented):
Concept makes sense and implementation is looking good. The styling of the indentation levels doesn't look quite right on my end. Perhaps this is still WIP? 

Screen Shot 2020-10-08 at 10 12 54 AM

AndrwM:
@philschanely Hmm, taking a look thanks!

philschanely (approved):
No concerns. May be unicornish but still useful in the system IMO. 🙌

joe-dakroub (approved):
No issues with these changes. Each update is evolving nicely! 👍

teenwolfblitzer (commented):
I'm seeing some odd behavior in Firefox. Chromium and Safari look great, so I'm not entirely sure why the layout isn't rendering:

Screen Shot 2020-10-12 at 11 32 57 AM

AndrwM:
Thanks @monkeypox8! I suspect that this is the offender:
https://github.com/Kajabi/sage/pull/442/files#diff-f51a6ce9087c83258c0a6db62bcf833dR27-R39

Will take a look 👍

PR #445, opened 2020-10-12
Table updates

philschanely:
## Description

Updates table styling per updated Sage table specs and adds Rails Table component.

### Screenshots

Screen Shot 2020-10-12 at 11 57 11 AM


lib/sage-frontend/stylesheets/system/patterns/elements/_table.scss
```diff
@@ -1,80 +1,101 @@
...
 
 ================================================== */
-
-$-sage-table-row-color-stripe: sage-color(grey, 100);
-$-sage-table-row-color-hover: sage-color(grey, 200);
-$-sage-table-cell-focus: sage-color(charcoal);
-$-sage-table-cell-truncate-width: 6.5em;
-$-sage-table-sort-indicator-width: rem(5px);
-$-sage-table-sort-indicator-direction: rem(7px);
+// Borders and sizing
```
teenwolfblitzer, line 8: 🌟

joe-dakroub (approved):
A lot of good cleanup, bug fixes and re-organization. No concerns with the code here. 👍 

One thing I noticed on my display, is that the striping is very subtle. Could just be my aging eyes, but worth mentioning.

teenwolfblitzer (commented):
👍 Looking much tidier, great job on the cleanup! Looks like adjustments may be in order for the responsive table breakpoints, which seem out of alignment with the sidebar/panel sizing:

table-800px

philschanely:
> A lot of good cleanup, bug fixes and re-organization. No concerns with the code here. 👍
> 
> One thing I noticed on my display, is that the striping is very subtle. Could just be my aging eyes, but worth mentioning.

Thanks for calling that out, @joedakroub  and I agree. Its also not in our newest specs yet, but I was not sure I should remove it either. I assume as we move forward we'll get clarification from Design if its desired.

philschanely:
@monkeypox8 Hey thanks for catching that! I'd adjusted the responsive effect because the fade was also covering the borders that were supposed to go "full bleed" inside panels now. This goofed things up! 🤦 

The revisions I just posted should patch it. I'm basically converting the borders under head and above footer to be absolutely positioned pseudo-elements that can therefore stack above the fade. What do you think?

lib/sage-frontend/stylesheets/system/patterns/elements/_table.scss
```diff
@@ -1,80 +1,107 @@
...
-      padding-right: sage-spacing(sm) / 2;
-    }
+  // Adds borders in these positions that are also stacked above the
+  // responsive scroll fade when it is enabled.
+  thead::after,
+  tfoot::before {
+    content: "";
+    display: block;
+    position: absolute;
+    z-index: 2;
```
teenwolfblitzer: Could this make use of the `sage-z-index()` helper?

lib/sage-frontend/stylesheets/system/patterns/elements/_table.scss
```diff
@@ -194,17 +278,18 @@ $-sage-table-sort-indicator-direction: rem(7px);
     position: absolute;
     top: 0;
     right: 0;
+    z-index: 1;
```
teenwolfblitzer: Same as the above comment re: `sage-z-index()` helper

teenwolfblitzer (approved):
👍 LGTM, looks 😎! Left a minor nitpick about using the z-index helper in place of values. I see a couple other files where that could be cleaned up, so it can wait to be addressed in a follow-up.

PR #446, opened 2020-10-13
Misc adjustments Post 1.15.0

philschanely:
## Description

A variety of small changes are included in this PR:

- Updated color palette to latest from Design.
- Added modifiers for `sage-row` to allow for spacing in stage, panel, and card contexts when row is used there.
- Added an `align-right` option for table cells.
- Patched a bug in Breadcrumbs Rails component.

No breaking changes to previous release.


pixelflips (approved):
👍  - LGTM +1

PR #450, opened 2020-10-15
Alert Updates

pixelflips:
## Description
Updates alerts to align with updated design specs and includes new action links.

**Note:** Icon updates will come in future PR.

### Screenshots

|  before  |  after  |
|--------|--------|
| Screen Shot 2020-10-15 at 12 21 54 PM|Screen Shot 2020-10-15 at 12 22 15 PM|

### DIFF
--- a/lib/sage-frontend/stylesheets/system/patterns/objects/_alert.scss
+++ b/lib/sage-frontend/stylesheets/system/patterns/objects/_alert.scss
@@ -40,11 +40,10 @@ $-alert-colors: (
 }
 
 .sage-alert__copy {
-  padding-right: rem(26px);
+  @include sage-grid-stack();
 }
 
 .sage-alert__title {
-  margin-bottom: sage-spacing(xs);
   @extend %t-sage-body-semi;
   @each $name, $color in $-alert-colors {
     .sage-alert--#{$name} & {
@@ -72,9 +71,6 @@ $-alert-colors: (
 
 .sage-alert__close {
   overflow: hidden;
-  position: absolute;
-  top: sage-spacing(sm);
-  right: sage-spacing(sm);
   height: rem(27.65px);
   width: rem(27.65px);
   padding: 0;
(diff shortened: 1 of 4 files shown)


lib/sage-frontend/stylesheets/system/patterns/objects/_alert.scss
```diff
@@ -12,23 +12,28 @@ $-alert-colors: (
 );
 
 .sage-alert {
-  display: flex;
+  display: grid;
```
philschanely, line 15: 🙌 here! For the `__copy` element you could consider implementing `@mixing sage-grid-stack`, which uses the 8px gap for a stack of elements. Then you'd not need any margin below the individual items and it would gracefully adjust with no `actions` there too.

lib/sage_rails/app/views/sage_components/_sage_alert.html.erb
```diff
@@ -1,8 +1,13 @@
...
   <i class="sage-alert__icon <%= component.icon_name %>" aria-hidden="true"></i>
   <div class="sage-alert__copy">
     <h3 class="sage-alert__title"><%= component.title %></h3>
     <p class="sage-alert__desc"><%= component.desc %></p>
+    <% if content_for? :sage_alert_actions %>
+      <div class="sage-alert__actions">
+        <%= content_for :sage_alert_actions %>
+      </div>
+    <% end %>
     <% if component.dismissable %>
```
philschanely: If you move this outside of `copy` it would then conform to the grid and auto-place in the upper right so you could likely remove the padding on copy and the absolute positioning. 🤔

philschanely (approved):
Great work here. A few things suggested below to hone further if desired.

lib/sage-frontend/stylesheets/system/patterns/objects/_alert.scss
pixelflips: Updated! Thanks for that!

lib/sage_rails/app/views/sage_components/_sage_alert.html.erb
pixelflips: @philschanely ❤️

PR #451, opened 2020-10-15
A few more misc patches post 1.15

philschanely:
## Description

- Add flexbox to `property` and a distinct `__value` element within to wrap text that affords more possibilities with the component.
- Revise breakpoints for layouts in card and panel 1024 instead of ~1200
- Add some missed styling for card lists to match same treatment as in Panels
- Add vertical centering to button groups
- Add some top and bottom spacing modifiers for cards and panels for instances where spacing should collapse.

No breaking changes.


joe-dakroub (approved):
Changes look good; no issues. Nice updates! 👍

PR #455, opened 2020-10-16
Alert Follow Up

pixelflips:
## Description
Added conditions for title and description so the markup doesn't appear empty when these sections have no content provided.

In addition, removed border property from the alert stylesheet that is no longer needed in the updated design.

**Note:** this update will cause no visual changes.


joe-dakroub (approved):
Changes look good to me. Thanks Phillip! 👍

philschanely (approved):

PR #460, opened 2020-10-19
Button updates

philschanely:
## Description

Updates buttons to spec along with implementations that needed to be updated as a result.

Of note:

- Buttons now only have `primary`, `secondary`, and `danger` color styles: `tertiary` is dropped.
- Buttons no longer have `size` variations. Instead there is just the default button, the `Subtle` button, and a `small` option for Subtle buttons that just makes smaller font size.

**Introduces breaking changes.** Products PR: https://github.com/Kajabi/kajabi-products/pull/16118

### Screenshots

Screen Shot 2020-10-19 at 1 10 19 PM

### DIFF
--- a/app/views/application/_assistant.html.erb
+++ b/app/views/application/_assistant.html.erb
@@ -11,34 +11,35 @@
   <div class="sage-assistant__body">
   </div>
   <div class="sage-assistant__actions">
-    <%= link_to "💎 Gem v#{$SAGE_VERSION_GEM}",
-      $SAGE_VERSION_GEM_URL,
-      class: "sage-label sage-label--draft sage-spacer-right-2xs",
-      target: "_blank",
-      rel: "noopener noreferrer",
-      title: "Open SageRails gem source"
-    %>
-    <%= link_to "📦 Package v#{$SAGE_VERSION_FRONTEND}",
-      $SAGE_VERSION_FRONTEND_URL,
-      class: "sage-label sage-label--draft sage-spacer-right-2xs",
-      target: "_blank",
-      rel: "noopener noreferrer",
-      title: "Open Sage frontent package source"
-    %>
-    |
-    <%= link_to "Release Notes",
-      $SAGE_RELEASE_URL,
-      class: "sage-btn sage-btn--small sage-btn--tertiary",
-      target: "_blank",
-      rel: "noopener noreferrer",
-      title: "Github release notes"
-    %>
-    <%= link_to "Changelog",
-      "https://github.com/Kajabi/sage/releases",
-      class: "sage-btn sage-btn--small sage-btn--secondary",
-      target: "_blank",
-      rel: "noopener noreferrer",
-      title: "Github changelog"
-    %>
+    <div class="sage-btn-group sage-btn-group--gap-md">
+      <%= link_to "💎 v#{$SAGE_VERSION_GEM}",
+        $SAGE_VERSION_GEM_URL,
+        class: "sage-btn sage-btn--subtle sage-btn--secondary",
+        target: "_blank",
+        rel: "noopener noreferrer",
+        title: "Open SageRails gem source"
+      %>
+      <%= link_to "📦 v#{$SAGE_VERSION_FRONTEND}",
+        $SAGE_VERSION_FRONTEND_URL,
+        class: "sage-btn sage-btn--subtle sage-btn--secondary",
+        target: "_blank",
+        rel: "noopener noreferrer",
+        title: "Open Sage frontent package source"
+      %>
+      <%= link_to "Release Notes",
+        $SAGE_RELEASE_URL,
+        class: "sage-btn sage-btn--subtle sage-btn--secondary",
+        target: "_blank",
+        rel: "noopener noreferrer",
+        title: "Github release notes"
+      %>
+      <%= link_to "Changelog",
+        "https://github.com/Kajabi/sage/releases",
+        class: "sage-btn sage-btn--subtle sage-btn--secondary",
+        target: "_blank",
+        rel: "noopener noreferrer",
+        title: "Github changelog"
+      %>
+    </div>
   </div>
 </div>
--- a/app/views/examples/elements/label/_preview.html.erb
+++ b/app/views/examples/elements/label/_preview.html.erb
@@ -12,7 +12,10 @@ colors = [
 <% colors.each do | color | %>
   <%= sage_component SageLabel, {
     color: color,
-    value: color.capitalize
+    value: %(
+      #{color.capitalize}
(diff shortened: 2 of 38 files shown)


lib/sage-frontend/stylesheets/system/core/_variables.scss
```diff
@@ -62,36 +62,6 @@ $sage-form-element-spacing: sage-spacing(sm) !default;
 // BUTTONS
 // ==================================================
 
-// Base styles
```
philschanely, line 65: Moved into Button where they're used.

lib/sage-frontend/stylesheets/system/patterns/elements/_status_icon.scss
```diff
@@ -11,29 +11,27 @@
   display: inline-flex;
   align-items: baseline;
   position: relative;
-  padding: $sage-iconbtn-padding;
-  color: $sage-btn-tertiary-text-color;
-  border-radius: sage-border(radius);
+  color: sage-color(charcoal, 200);
```
philschanely, line 14: Revisions here sync up with the new Subtle button styles.

lib/sage-frontend/stylesheets/system/patterns/objects/_outline_item.scss
```diff
@@ -21,7 +21,7 @@ $-collapse-breakpoint-key: md-max;
   display: grid;
   align-items: center;
   grid-template-rows: repeat(auto-fit, auto);
-  grid-column-gap: sage-spacing(xs);
+  grid-column-gap: sage-spacing(sm);
```
philschanely, line 24: Changes to the button styles affected spacings here so this just syncs up with mocks.

lib/sage-frontend/stylesheets/system/patterns/objects/_sortable.scss
```diff
@@ -13,15 +13,15 @@
   display: flex;
   flex-direction: row;
   align-items: center;
-  gap: sage-spacing(2xs);
-  padding: sage-spacing(2xs);
+  gap: sage-spacing(sm);
```
philschanely: Also here new button styles impacted spacing so this syncs things up with the mocks.

lib/sage-frontend/stylesheets/system/core/_icons.scss
```diff
@@ -59,10 +59,57 @@ $sage-icon-li-margin-right: 0.4em !default;
     }
   }
 
+
   // Other icons generated here
+  
+  @include button-icon-generator($icon-name, $icon-code, only);
```
philschanely, line 65: Testing a theory to help our compilation run faster I consolidated all the loops that generate icon variants into this one spot so one loop covers all the uses.

app/views/examples/objects/page_heading/_preview.html.erb
```diff
@@ -2,14 +2,14 @@
 
 <%= sage_component SagePageHeading, {
```
philschanely, line 3: This Page Heading component now makes use of the Subtle buttons for the toolbar and the `raised` prop on the Secondary button to ensure it appears to float like the primary button the grey background.

app/views/examples/objects/outline_item/_preview.html.erb
```diff
@@ -6,19 +6,22 @@
   status: sage_component(SageLabel, { value: "Draft", color: "draft" }),
```
philschanely, line 6: Outline Items can now make use of the smaller Subtle buttons

app/views/examples/objects/catalog_item/_preview.html.erb
```diff
@@ -5,20 +5,23 @@
     href: "#"
   } do %>
     <%= sage_component SageButton, {
-      style: "tertiary",
-      size: "tiny",
+      style: "secondary",
```
philschanely, line 8: Catalog items can also make use of the small Subtle buttons.

app/views/examples/elements/label/_preview.html.erb
```diff
@@ -12,7 +12,10 @@ colors = [
 <% colors.each do | color | %>
   <%= sage_component SageLabel, {
     color: color,
-    value: color.capitalize
+    value: %(
```
philschanely, line 15: This is a temporary demo of using buttons within labels and will be removed before merge. FYI @QuintonJason

app/views/application/_assistant.html.erb
```diff
@@ -11,34 +11,35 @@
...
-      title: "Github release notes"
-    %>
-    <%= link_to "Changelog",
-      "https://github.com/Kajabi/sage/releases",
-      class: "sage-btn sage-btn--small sage-btn--secondary",
-      target: "_blank",
-      rel: "noopener noreferrer",
-      title: "Github changelog"
-    %>
+    <div class="sage-btn-group sage-btn-group--gap-md">
```
philschanely, line 14: These updates revise the button styles to reverse inside the assistant.

philschanely:
@AndrwM This PR touches a few components you've worked on recently and introduces breaking changes. I have a PR started (linked above) that provides the fixes but I appreciate your eyes as well if you have time.

lib/sage-frontend/stylesheets/system/core/_icons.scss
joe-dakroub: Great thinking!

joe-dakroub (approved):
These changes look good to me. Thank you for the helpful, inline comments too Phil! 👍

pixelflips (approved):
Nice work! LGTM 👍

lib/sage-frontend/stylesheets/system/core/_mixins.scss
```diff
@@ -174,30 +246,33 @@
...
 
   &::after {
+    $-offset-block: $outline-width * 2 + $outline-offset-block * 2;
+    $-offset-inline: $outline-width * 2 + $outline-offset-inline * 2;
+
     content: "";
     position: absolute;
     left: 50%;
     top: 50%;
     transform: translate3d(-50%, -50%, 0) scale(0.94);
```
teenwolfblitzer, line 264: Something I've been noticing on the buttons for a while in Chrome is that the transform `scale(0.94)` introduces a small amount of rounding error as the focus outlines are drawn. This becomes a little more obvious on the reversed version, where the right part of the outline seems to be 'clipped'.

A quick fix I was playing with adds `0.5px` width to the `$-offset-inline` value:

| rounding error | added width |
|-----|-----|
|before|after|

Just a thought, because I'm not so sure that adding width to the pseudo-element is the ideal way to fix it. Also, this doesn't seem to occur in Firefox or Safari, so it's very much a nice-to-have.

teenwolfblitzer (approved):
👍  🔥  Great work! One non-blocking item for future follow-up/further investigation.

PR #477, opened 2020-10-20
[Rails] Fix Broken :html_attributes Concatenation

AndrwM:
## Description
Rails view was considering `generated_html_attributes` as "unsafe html". I tried a few workarounds but had some trouble getting this fixed. I settled on explicitly calling `html_safe` in each instance of the usage.

Sorry about the noisy diff here, all-in-all I think this explicitness is probably best anyway. 👍


pixelflips (approved):
👍  +1 - Thanks for fixing!

PR #478, opened 2020-10-21
Update warning alert icon color

pixelflips:
## Description
Small update to change warning icon color from yellow to orange in alerts. 
Missed this change in the recent updates.

### DIFF
--- a/lib/sage-frontend/stylesheets/system/patterns/objects/_alert.scss
+++ b/lib/sage-frontend/stylesheets/system/patterns/objects/_alert.scss
@@ -7,7 +7,7 @@
 $-alert-colors: (
   info: primary,
   success: sage,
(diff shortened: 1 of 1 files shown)


AndrwM (approved):
+1! ⚠️

QuintonJason (approved):
LGTM! I only suggestion would be to create a map for the icon colors, `$-alert-icon-colors`. Then we'd loop over that map on line 64

pixelflips:
> LGTM! I only suggestion would be to create a map for the icon colors, `$-alert-icon-colors`. Then we'd loop over that map on line 64

@QuintonJason 🙌  - thanks so much for the tip! I knew I was overthinking it and I appreciate the solid advice!

PR #481, opened 2020-10-21
Tooltips: allow children to rollover

AndrwM:
## Description
Previously when rolling over a child element of a parent-bound tooltip the tooltip would disappear. This fixes that.

Mouseleave docs:
https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseleave_event

Mouseout docs:
https://developer.mozilla.org/en-US/docs/Web/API/Element/mouseout_event

## Test notes
Test to ensure tooltips rollover

### Steps for testing
1. Review app
2. Check "created at" tooltip at: `https://www.kajabi.test:3000/admin/sites/:id/products`

## Related
BUILD-323

### DIFF
--- a/lib/sage-frontend/javascript/system/tooltip.js
+++ b/lib/sage-frontend/javascript/system/tooltip.js
@@ -13,13 +13,13 @@ Sage.tooltip = (function() {
   // ==================================================
 
   function init(el) {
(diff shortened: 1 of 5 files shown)


pixelflips (approved):
LGTM! Nice work! 🖱️

philschanely (approved):
Nice thinking.

PR #482, opened 2020-10-21
SageSearch: --contained modifier

AndrwM:
## Description
Add `--contained` for incoming Search "Toolbar" pattern.

### Screenshots
screencast 2020-10-21 12-43-32

### Steps for testing
1. Check review app

## Related
BUILD-323

### DIFF
--- a/lib/sage-frontend/stylesheets/system/patterns/elements/_search.scss
+++ b/lib/sage-frontend/stylesheets/system/patterns/elements/_search.scss
@@ -4,19 +4,11 @@
 
 ================================================== */
 
-
 $-search-icon: "::before";
-$-search-focus-line: "::after";
 
 .sage-search {
   position: relative;
   background-color: sage-color(white);
-  border-bottom: 1px solid sage-color(grey, 300);
-
-  &.sage-search--large {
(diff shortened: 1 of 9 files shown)


lib/sage-frontend/stylesheets/system/patterns/elements/_search.scss
```diff
@@ -10,8 +10,14 @@ $-search-focus-line: "::after";
 
 .sage-search {
   position: relative;
+  background-color: sage-color(white);
   border-bottom: 1px solid sage-color(grey, 300);
 
+  &.sage-search--large {
```
philschanely: @AndrwM is there something other than size that could distinguish this? There also doesn't seem to be any size variation happening. Since it adds a full border would something like "full-border" make sense instead? I'm also about to start working on the Panel Controls component and in that context perhaps "as-control" or just "control" would make sense?

AndrwM: This has been resolved via slack.

philschanely (approved):
😍 🏆

PR #484, opened 2020-10-21
Sage ProgressBar

AndrwM:
## Description
- Sage Progressbar that internally uses `<progress>`. (Custom appearance styles don't have enough cross-browser support to style the actual element.)
- Update SageCatalog card to be considered 'disabled' when a href isn't passed to it
- Update SageCatalog borders to match SagePanelListItem styles

**Initial comps:**
https://www.figma.com/file/TO8XSqt6tgscFcShhkWVdm/Products-Admin?node-id=2501%3A18351

**My designer approved the below:**
image

### Steps for testing
1. Ensure proper a11y for progressbar in docs

## Related
BUILD-323

### DIFF
--- a/lib/sage_rails/app/views/sage_components/_sage_progress_bar.html.erb
+++ b/lib/sage_rails/app/views/sage_components/_sage_progress_bar.html.erb
@@ -1,9 +1,21 @@
+<% content = component.label ? "#{component.label}: " : "" %>
+<% content << "#{component.percent}%&nbsp;progress" %>
+
 <div
   class="sage-progress-bar"
+  data-js-tooltip="<%= content %>"
   <%= component.generated_html_attributes.html_safe %>
 >
-  <progress class="sage-progress-bar__element" max="100" value="<%= component.percent %>">
-    Progress: <%= component.percent %>%
+  <progress
+    class="sage-progress-bar__element"
+    aria-valuemax="100"
+    max="100"
+    aria-valuemin="0"
+    value="<%= component.percent %>"
+    aria-valuenow="<%= component.percent %>"
+    aria-valuetext="<%= content %>"
+  >
+    <%= content %>
   </progress>
-  <div class="sage-progress-bar__value" style="width: <%= component.percent %>%;">&nbsp;</div>
+  <div class="sage-progress-bar__value" style="width: <%= component.percent %>%;"></div>
 </div>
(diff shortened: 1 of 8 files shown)


lib/sage_rails/app/views/sage_components/_sage_progress_bar.html.erb
```diff
@@ -0,0 +1,9 @@
+<div
+  class="sage-progress-bar"
+  <%= component.generated_html_attributes.html_safe %>
+>
+  <progress class="sage-progress-bar__element" max="100" value="<%= component.percent %>">
+    Progress: <%= component.percent %>%
+  </progress>
+  <div class="sage-progress-bar__value" style="width: <%= component.percent %>%;">&nbsp;</div>
```
teenwolfblitzer: Is the `&nbsp;` character necessary here?

If so, then I think `sage-progress-bar__value` would need `aria-hidden="true"`. I'm not 100% certain because voiceover doesn't pick up the element, but Firefox does see it as sectioning content because of the non-breaking-space.

lib/sage_rails/app/views/sage_components/_sage_progress_bar.html.erb
```diff
@@ -0,0 +1,9 @@
+<div
+  class="sage-progress-bar"
+  <%= component.generated_html_attributes.html_safe %>
+>
+  <progress class="sage-progress-bar__element" max="100" value="<%= component.percent %>">
```
teenwolfblitzer: I believe a corresponding `label` or an `aria-label` is needed for a `progress` element, we'll want to describe what it's measuring.

teenwolfblitzer: Not a huge deal, but Safari being Safari doesn't display the animation, and instead "pops" in the element once the animation sequence is complete. Last time I had to custom style a progress bar like this was 2+ years ago, nice to know Safari hasn't changed at all since then 💩

teenwolfblitzer (commented):
Looking great, love the animation! Left a couple of comments to consider around accessibility. Safari doesn't display the animation, but that seems to be a minor detail.

lib/sage_rails/app/views/sage_components/_sage_progress_bar.html.erb
AndrwM: I don't think it is :) Old habits. Thanks!

AndrwM:
Updated based on @monkeypox8's feedback ✨

PR #490, opened 2020-10-22
Sage Label Nested Buttons

AndrwM:
## Description

This PR updates the interactive label.  The interactive label now calls for two icons. The second icon will also be a clickable button. The clickable button is a `SageButton` component that has `icon: { style: "only", name: "caret-down" }`

### Screenshots
|  before  |  after  |
|--------|--------|
|interactive-tags-before|interactive-tags-after|

### Steps for testing
1. Visit the Sage Label page: http://localhost:4000/pages/element/label

### DIFF
--- a/lib/sage-frontend/stylesheets/system/patterns/elements/_label.scss
+++ b/lib/sage-frontend/stylesheets/system/patterns/elements/_label.scss
@@ -88,7 +88,7 @@ $-label-inset-border: 0 0 0 1px inset;
 
     &:hover {
       .sage-label__value {
-        background-color: sage-color-combo($-color-name, default, background-accent)
+        background-color: sage-color-combo($-color-name, default, background-accent);
       }
 
       .sage-btn::before {
@@ -97,9 +97,9 @@ $-label-inset-border: 0 0 0 1px inset;
     }
 
     .sage-btn {
+      display: flex;
       width: rem(24px);
       min-height: rem(24px);
-      display: flex;
       justify-content: center;
       border-radius: 0 sage-border(radius) sage-border(radius) 0;
 
--- a/lib/sage-frontend/stylesheets/system/patterns/elements/_label.scss
+++ b/lib/sage-frontend/stylesheets/system/patterns/elements/_label.scss
@@ -36,6 +36,10 @@ $-label-inset-border: 0 0 0 1px inset;
   position: absolute;
   right: sage-spacing(xs);
   margin: auto 0;
+
+  .sage-btn::before {
+    font-size: rem(8px);
+  }
 }
 
 .sage-label--truncate {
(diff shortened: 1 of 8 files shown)


QuintonJason:
Need to update the color of the icon only button. Using `sage-color-combo` but color is not correct

lib/sage-frontend/stylesheets/system/patterns/elements/_label.scss
```diff
@@ -9,17 +9,39 @@ $-label-padding: 0 sage-spacing(xs) !default;
...
   border: unset;
+  outline: none;
+
+  &:hover,
+  &:focus,
+  &:active {
+    outline: none;
+  }
+
+  .sage-btn {
```
philschanely: Could we please move this to `_button` like this:
```scss
.sage-btn {
  ...
  .sage-label__value & { ... }
}
```

lib/sage-frontend/stylesheets/system/patterns/elements/_label.scss
```diff
@@ -28,41 +50,99 @@ $-label-inset-border: 0 0 0 1px inset;
...
-      &:focus,
-      &:active {
-        background-color: inherit;
-        box-shadow: $-label-inset-border transparent;
+    .sage-btn {
+      display: flex;
+      justify-content: center;
+      position: absolute;
+      right: 0;
+      width: rem(24px);
```
philschanely: Could these be stored in a local variable?

lib/sage-frontend/stylesheets/system/patterns/elements/_label.scss
```diff
@@ -28,41 +50,99 @@ $-label-inset-border: 0 0 0 1px inset;
...
+        color: sage-color-combo($-color-name, default, foreground-accent);
       }
     }
 
-    .sage-btn & {
-      &:focus,
-      &:active {
-        background-color: inherit;
-        box-shadow: $-label-inset-border transparent;
+    .sage-btn {
```
philschanely: Can this be moved to the `_button` same as above? The hope is to maintain all styles for a given thing in that thing's file. I realize this may make things a little more complex in that file but I believe have things consolidated like that helps changes and their scope be easier to digest.

lib/sage-frontend/stylesheets/system/tokens/_color_combos.scss
```diff
@@ -10,8 +10,10 @@ $sage-color-combos: (
   draft: (
```
philschanely, line 15: Seeing this now I believe you could bring over those new colors to this file to assign within and thus make them a little more globally available.

philschanely (commented):
Fabulous work here, @QuintonJason ! you've done much elegant work that definitely improves this component. I have some organizational comments but also have a question related to the outer affordance I'll ask in the squad channel and copy you.

lib/sage-frontend/stylesheets/system/patterns/elements/_label.scss
```diff
@@ -28,41 +50,99 @@ $-label-inset-border: 0 0 0 1px inset;
...
+        height: rem(24px);
+        border-radius: 0 sage-border(radius) sage-border(radius) 0;
+      }
+
+      &:focus {
+        @include sage-focus-outline--update-color( sage-color-combo($-color-name, default, foreground-accent) );
+      }
+
+      &.sage-btn--tag::before {
+        font-size: rem(16px);
```
teenwolfblitzer: Small detail, we can use `sage-font-size()` here

teenwolfblitzer (commented):
Aside from the scoping items @philschanely touched on, no concerns here. Looking/working great!

One item I'd like to get more context on later is touch target sizes. On some elements it's starting to get a little _too_ small for both mobile and accessibility considerations.

PR #492, opened 2020-10-23
Panel controls

philschanely:
Panel controls allow us to control a table of items and provide a variety of flexible options. This is scaffolded to pair with a Rails table, but a similar build out in React will follow.


app/views/examples/objects/panel_controls/_preview.html.erb
```diff
@@ -0,0 +1,355 @@
+<%
+tabs = sage_component(SageTabs, {
```
philschanely, line 2: Frontmatter here sets up a few aspects that are re-used in the examples below.

app/views/examples/objects/panel_controls/_preview.html.erb
```diff
@@ -0,0 +1,355 @@
...
+  show_bulk_actions: true,
+  show_expand_collapse: true,
+  show_pagination: true,
+  show_sort: true,
+  bulk_action_items: actions_items,
+  sort_items: sort_items,
+} %>
+
+<h4>Sample Custom Toolbar</h4>
+<%= sage_component SagePanelControls, {} do %>
```
philschanely, line 138: Places like the main People page will use a customized toolbar like this one.

app/views/examples/objects/panel_controls/_preview.html.erb
```diff
@@ -0,0 +1,355 @@
...
+Items provided to the `bulk_action_items`
+should set the `data-js-action` attribute
+with the name of an action to be exposed to the responder.
+This value is passed to the `detail` of the custom event
+through the `action` property.
+Any additional properties may be provided as data attributes
+and accessible on the custom event's `target.dataset` property.
+), use_sage_type: true) %>
+
+<script type="text/javascript">
```
philschanely, line 307: This bit shows what some pairing in the app could look like. It is allowing the panel controls to control selections in this list.

app/views/examples/objects/panel_controls/_preview.html.erb
```diff
@@ -0,0 +1,355 @@
...
+  <% content_for :sage_panel_controls_tabs do %>
+    <%= tabs %>
+  <% end %>
+  <% content_for :sage_panel_controls_toolbar do %>
+    <%= search_filter_toolbar %>
+    <%= dropdown %>
+  <% end %>
+<% end %>
+
+<%= sage_component SageTable, {
```
philschanely, line 164: This demo table pairs with the Kitchen sink example to show some of the potential for pairing. The `data-` props below could be controlled by server-side code and provide information to the panel controls.

teenwolfblitzer (commented):
This is looking great so far! I'm wondering what the expected layout for mobile should look like. Right now, the controls break out of the surrounding container at anything less than `1024px`:

philschanely:
@monkeypox8 Great catch. Responsiveness for these more complex toolbars needs further attention in a follow up and when the custom toolbar is implemented. I've at least patched to ensure the default ones built into the panel controls wrap if a little awkward for now.

lib/sage_rails/app/views/sage_components/_sage_table.html.erb
```diff
@@ -26,11 +26,15 @@ is_responsive = component.responsive.present? ? component.responsive : true
         <tr>
           <% component.headers.each do | header | %>
             <th
+              class="
```
teenwolfblitzer, line 30: As a follow-up item, I wonder if it'd be possible to add an option for a visually-hidden `th` when table heading text is not intended to be shown. As this demo shows, an empty heading is sometimes used for the checkbox column in the app.

lib/sage_rails/app/views/sage_components/_sage_panel_controls.html.erb
```diff
@@ -0,0 +1,121 @@
...
+      <%= content_for :sage_panel_controls_toolbar %>
+    </div>
+  <% end %>
+
+  <% if show_default_controls %>
+    <div class="sage-panel-controls__default-controls">
+      <% if component.show_bulk_actions %>
+        <div class="sage-panel-controls__bulk-actions">
+          <%= sage_component SageCheckbox, { 
+            id:"c1",
```
teenwolfblitzer: Small cleanup item for the docs:  `c1` is being used as an id in several places, preventing labels from being associated with their controls. This is especially true on the objects main page, where the icon list gets toggled instead.

app/views/examples/objects/panel_controls/_preview.html.erb
```diff
@@ -0,0 +1,383 @@
...
+  let selectedItemsCount = null;
+
+  // Panel Controls object
+  let panelControls = null;
+
+  // Functions
+
+  const setupPanelControls = () => {
+    table = document.getElementById('demo-table');
+    items = table.querySelectorAll('tr');
```
teenwolfblitzer, line 331: I'm seeing an error in the console for the demo script here; looks like the `demo-table` id isn't getting output on the table.

teenwolfblitzer (approved):
👍  LGTM, looking forward to seeing this in action! Couple of minor items to ponder for our own internal (docs) benefit, nothing blocking. Great work!

joe-dakroub (approved):
What a terrific set of additions! Looking great.

I do see the same JS error that @monkeypox8 mentioned, but otherwise this is in good shape. Really nice work Phil! 👍

PR #493, opened 2020-10-25
Add markdown

philschanely:
## Description

This is the first of a few documentation experience updates I've been working on. 

- Adds the [Redcarpet Gem](https://github.com/vmg/redcarpet) and a markdown helper to use markdown for documenting our system.
  - `md(text <string>, use_sage_type <boolean>)` allows us to provide a string that is parsed as Markdown and optionally wrap it with a `div.sage-type`.
  - Prep for improved rendering of code blocks in docs (See Sandbox for a peak)
- Implements this approach throughout the system for documentation blocks.

NOTE: Temporary code blocks in sandbox will be removed before merge.


philschanely:
Sorry this is so large, y'all. Very little is actually changed in the docs themselves beyond implementing Markdown and doing some style tidying for the previews.

app/helpers/objects_helper.rb
```diff
@@ -12,16 +12,6 @@ module ObjectsHelper
   def sage_objects
     [
       # Sage Generated Objects
-      {
```
philschanely, line 15: We accidentally had a duplicate entry for this.

app/views/examples/elements/_element_preview.html.erb
```diff
@@ -14,7 +14,5 @@
       <% end %>
     </div>
   </div>
-  <div class="sage-type">
```
philschanely, line 17: We can remove this now, allowing the Panel spacing to clean up somethings and then the new `md` function to provide this container when its actually needed throughout the docs.

app/views/examples/elements/copy_text/_rules_do.html.erb
```diff
@@ -1,4 +1,6 @@
-<ul>
-  <li>Use inline <code>&lt;b&gt;</code> tag to add bolding to select portions of these blocks.</li>
-  <li>Provide a local max-width as desired in order to limit overflowing text. Truncation is setup and ready for this by default on the inline Copy Text, but can be turned on for the Copy Text Card with <code>truncate_contents</code> set to <code>true</code>.</li>
-</ul>
+<%= md("
```
philschanely, line 1: The "rules" pages are simply a call to the `md` function now. future may convert them to `.md` files altogether.

Recommending using the `"` for wrapping the string to be parsed, but we just have to watch to escape actual `"` that are needed within, or use one of the alternate string enclosures.

app/views/pages/index.html.erb
```diff
@@ -1,16 +1,88 @@
...
+
+      Rails does many things 1very well and gives us the ability to move quickly
+      and solve complex problems with very simple, tried and true, code solutions.
+      With that said, sometimes we come across a problem that we want to solve
+      that the standard rails approach will just not be enough
+      (think complex interactions with many versions of state ).
+      In this case we move to React as our default approach
+      to solving complex, Javascript heavy problems
+
+Some HTML:
```
philschanely, line 27: Just some demo code blocks here to show Rouge at work. Will remove before merge.

lib/sage-frontend/stylesheets/docs/_rouge_theme.scss
```diff
@@ -0,0 +1,323 @@
+// background
```
philschanely, line 1: Initial crack at styling the Rouge code blocks. Suggestions or forks welcome!

app/views/examples/elements/checkbox/_preview.html.erb
```diff
@@ -1,53 +1,55 @@
 <!-- Default -->
-<%= sage_component SageCheckbox, { 
-  id:"c1",
-  label_text: "Checkbox",
-  checked: false,
-  disabled: false,
-  has_error: false
-} %>
+<div class="sage-panel__stack">
```
AndrwM: Would it be better to use the `SagePanelStack` rails component here?

app/helpers/markdown_helper.rb
```diff
@@ -0,0 +1,36 @@
+module MarkdownHelper
+  require "redcarpet"
+  require "rouge"
+  require "rouge/plugins/redcarpet"
+
+  class HTML < Redcarpet::Render::HTML
+    include Rouge::Plugins::Redcarpet
+  end
+
+  def md(text, use_sage_type=false)
```
AndrwM: Could we update this argument to be key/value? This will allow us to add other named args in the future and is more explicit about what the second arg is (rather than just `true`)

```
def md(text, use_sage_type: false)
  ...
end
```
```
md(..., use_sage_type: true)
```

app/helpers/markdown_helper.rb
```diff
@@ -0,0 +1,36 @@
...
+      # strikethrough: true,
+      # superscript: true,
+      # lax_spacing: true,
+    }
+ 
+    markdown = Redcarpet::Markdown.new(renderer, extras)
+    if use_sage_type
+      raw "<div class=\"sage-type\">#{markdown.render(text)}</div>"
+    else
+      raw markdown.render(text)
```
AndrwM, line 32: Is it possible to use `.html_safe` rather than `raw` on this?

`raw` is bad practice because it doesn't escape javascript injection (ofc, we have no security concerns about rouge user input on Docs so not a big deal either way)

app/views/examples/elements/label/_preview.html.erb
```diff
@@ -9,52 +9,62 @@ colors = [
...
-<% colors.each do | color | %>
-  <%= sage_component SageLabel, {
-    color: color,
-    value: %(
-      #{color.capitalize}
-      <button class="sage-btn sage-btn--subtle sage-btn--secondary sage-btn--icon-only-remove"></button>
-    )
-  } %>
-<% end %>
+<div class="sage-panel__block">
```
AndrwM: Same here regard SagePanelBlock rather than html/css-class

AndrwM (commented):
Really awesome work!!!!

I'm confused by the usage of `sage-type` with the markdown helper but this is a _much needed_ improvement!

app/helpers/markdown_helper.rb
```diff
@@ -0,0 +1,36 @@
...
+  def md(text, use_sage_type=false)
+    render_options = {
+      # filer_html: true, 
+      # hard_wrap: true,#
+      link_attributes: { rel: 'nofollow' },
+      with_toc_data: true
+    }
+    renderer = HTML.new(render_options)
+    extras = {
+      autolink: true,
```
teenwolfblitzer: Could this be useful as an optional arg for the `md` helper? It's the only item in the "extras" right now that I could see needing more granular control.

app/helpers/markdown_helper.rb
teenwolfblitzer: I'm glad Andrew called this out, it did seem a little ambiguous what the argument was referring to in the markdown blocks.

teenwolfblitzer (approved):
👍  Really excited for this one. Echoing Andrew's comment about labelling the `use_sage_type` argument but otherwise looks great to me!

app/helpers/markdown_helper.rb
philschanely: Good call, Gents. Fixed.

app/helpers/markdown_helper.rb
philschanely: Thank you! Fixed.

app/helpers/markdown_helper.rb
philschanely: Yeah actually for now I think I'd like that turned off. We can then add in later if desired.

philschanely:
Thanks for the feedback, @AndrwM ! I've adjusted the rails stuff accordingly.

> I'm confused by the usage of `sage-type` with the markdown helper but this is a _much needed_ improvement!

This is to speed up wrapping a `sage-type` container around markdown text and enables us to remove the heavy-handed wrapper we had around previews. In upcoming PRs I hope to push on this more to provide better structure for our code examples.

PR #495, opened 2020-10-27
Inline and button "copy text" overflow fix

teenwolfblitzer:
## Description
Applies a `max-width` to prevent long strings of text (like URLs) from overflowing "copy text" element bounds. This occurs most often on mobile devices and in limited-width containers, like modals.

### Screenshots

|  before  |  after  |
|--------|--------|
|before|after|

### Steps for testing
1. Navigate to the "Copy text" element preview
2. Decrease the viewport size of your browser to simulate a mobile view
3. Verify that the text does not overflow past the container


philschanely (approved):
Nice!! Thank you!

joe-dakroub (approved):
Great catch. Looks good! 👍

PR #497, opened 2020-10-27
link icons and nesting

cmcfadzean:
## Description
Updates the nav items to accept icons and also lets them be nested so that we can conditionally render them in the future. 

### Screenshots
Screen Shot 2020-10-27 at 9 45 29 AM

### DIFF
--- a/lib/sage-frontend/stylesheets/system/patterns/objects/_nav.scss
+++ b/lib/sage-frontend/stylesheets/system/patterns/objects/_nav.scss
@@ -54,8 +54,8 @@ $-sage-nav-icon-size: rem(20px);
 }
 
 .sage-nav__link--sub.sage-nav__link--active {
-  background: rgba($-sage-nav-color-background, 0.06);
   color: sage-color(primary);
+  background: rgba($-sage-nav-color-background, 0.06);
 }
 
 .sage-nav__link-decorator {
--- a/lib/sage-frontend/stylesheets/system/patterns/objects/_nav.scss
+++ b/lib/sage-frontend/stylesheets/system/patterns/objects/_nav.scss
@@ -16,6 +16,7 @@ $-sage-nav-icon-size: rem(20px);
 
 .sage-nav__list--sub {
   margin-left: rem(36px);
+
   .sage-nav__link--active {
     color: sage-color(primary);
     background: rgba($-sage-nav-color-background, 0.06);
@@ -58,7 +59,7 @@ $-sage-nav-icon-size: rem(20px);
 
 .sage-nav__link--active .sage-nav__icon {
   color: sage-color(primary);
-} 
+}
 
 .sage-nav__link-decorator {
   display: inline-block;
@@ -74,6 +75,7 @@ $-sage-nav-icon-size: rem(20px);
 
 .sage-nav__icon {
   margin-right: sage-spacing(xs);
+
   &::before {
     @include sage-icon-base(null, xl);
   }
--- a/lib/sage-frontend/stylesheets/system/patterns/objects/_nav.scss
+++ b/lib/sage-frontend/stylesheets/system/patterns/objects/_nav.scss
@@ -14,11 +14,19 @@ $-sage-nav-icon-size: rem(20px);
   list-style: none;
 }
 
+.sage-nav__list--sub {
+  margin-left: rem(36px);
+  .sage-nav__link--active {
+    color: sage-color(primary);
+    background: rgba($-sage-nav-color-background, 0.06);
+  }
+}
+
 .sage-nav__link {
   display: flex;
   align-items: center;
   height: rem(32px);
-  margin-bottom: rem(4px);
+  margin-bottom: sage-spacer(2xs);
   padding: sage-spacing(2xs) sage-spacing(xs);
   color: $-sage-nav-color-text;
   border-radius: sage-border(radius);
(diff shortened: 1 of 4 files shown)


lib/sage_rails/app/views/sage_components/_sage_nav_link.html.erb
```diff
@@ -1,7 +1,9 @@
 <% active_class = (component.no_active.blank? && current_page?(component.link)) ? "sage-nav__link--active" : "" %>
 <% component_class = component.type ? "sage-nav__link--#{component.type}" : "" %>
-<%= link_to component.text,
-  component.link,
-  method: component.method,
-  class: "sage-nav__link #{active_class} #{component_class}"
-%>
+<%= link_to component.link, method: component.method, class: "sage-nav__link #{active_class} #{component_class}" do %>
+  <% if component.icon %>
+    <div class="sage-icon-<%= component.icon %> sage-nav__icon"></div>
```
pixelflips: Should this be an inline element, possibly an `i`? I believe that would better match icons used in other elements/objects. 

`<i class="sage-alert__icon sage-icon-info-circle" aria-hidden="true"></i>`

Also seems to be a lint error due to ordering of properties.

pixelflips (approved):
Left a small comment below, but nothing show-stopping. LGTM 👍

lib/sage_rails/app/views/sage_components/_sage_nav_link.html.erb
pixelflips: @cmcfadzean ignore the linting error mention, looks like you already addressed it.

lib/sage-frontend/stylesheets/system/patterns/objects/_nav.scss
```diff
@@ -44,12 +45,16 @@ $-sage-nav-color-background: sage-color(primary, 400);
...
 .sage-nav__link--sub {
-  padding-left: sage-spacing(sm);
-
   @extend %t-sage-body-small-med;
+  margin-left: 36px;
+}
+
+.sage-nav__link--active .sage-nav__icon {
+  color: sage-color(primary);
 }
```
AndrwM, line 62: Could we move this down to line 79 with a pattern like so:

```
.sage-nav__icon {
...
  .sage-nav__link--active & {
     ...
  }
}
```

This aligns with our CSS organization pattern in Sage.

lib/sage-frontend/stylesheets/system/patterns/objects/_nav.scss
```diff
@@ -44,12 +45,16 @@ $-sage-nav-color-background: sage-color(primary, 400);
...
 
 .sage-nav__link--sub {
-  padding-left: sage-spacing(sm);
-
   @extend %t-sage-body-small-med;
+  margin-left: 36px;
```
AndrwM: Rem conversion needed here with possible `sage-spacer` usage if 32px is okay with Design

lib/sage-frontend/stylesheets/system/patterns/objects/_nav.scss
```diff
@@ -17,7 +18,7 @@ $-sage-nav-color-background: sage-color(primary, 400);
   display: flex;
   align-items: center;
   height: rem(32px);
-  margin-bottom: rem(2px);
+  margin-bottom: rem(4px);
```
AndrwM: Could we use `sage-spacer(2xs)` here instead?

lib/sage-frontend/stylesheets/system/patterns/objects/_nav.scss
```diff
@@ -64,3 +69,13 @@ $-sage-nav-color-background: sage-color(primary, 400);
...
+
+.sage-nav__icon {
+  margin-right: sage-spacing(xs);
+  &::before {
+    width: $-sage-nav-icon-size;
+    height: $-sage-nav-icon-size;
+    font-size: $-sage-nav-icon-size;
+    line-height: rem(32px);
+  }
+}
```
AndrwM: We have system icon sizes predefined within `tokens/_icon.scss`, they can be accessed using the `sage-icon-base` mixin.

This is an example:

```
  &::before {
    @include sage-icon-base(null, xl);
  }
```

Rather than keying in a line-height alignment with a magic number, could we use `flex` to align these icons?

AndrwM (approved):
+1 Left some comments. Looking great!

lib/sage-frontend/stylesheets/system/patterns/objects/_nav.scss
philschanely: Agreed; if possible it would be good to stick to our standard spacers. Also could this possibly be put on the `.sage-nav__list--sub`—the `ul` containing the sub items rather than the links within list items? Seems to me that styling a single container removes this opinion from the detail elements.

lib/sage-frontend/stylesheets/system/patterns/objects/_nav.scss
```diff
@@ -44,12 +45,16 @@ $-sage-nav-color-background: sage-color(primary, 400);
...
   @extend %t-sage-body-small-med;
+  margin-left: 36px;
+}
+
+.sage-nav__link--active .sage-nav__icon {
+  color: sage-color(primary);
 }
 
-.sage-nav__link--active {
+.sage-nav__link--sub.sage-nav__link--active {
```
philschanely: Can this be just `.sage-nav__link--active`? If specifity demands otherwise, lets instead place it inside `.sage-nav__link--sub`:

```sass
.sage-nav__link--sub {
  ...

  &.sage-nav__link--active {
    ...
  }
}
```

philschanely (commented):
Looking good. Seconding some others' input and one additional ask 🙏

philschanely:
@cmcfadzean I'm not sure if its part of your goal with this PR but it seems there are a number of areas this component needs to be updated to align with latest design specs [here](https://www.figma.com/file/SCZt9jMd7OcTMrNsOFifDp/?node-id=9%3A153). This would imply some adjustments to spacings, coloring, and interactive states. It also may be that the Rails component needs its structure updated. Again, may not be what you're aiming for in this PR. :-D

lib/sage-frontend/stylesheets/system/patterns/objects/_nav.scss
cmcfadzean: I believe that because the icons are 20px to get the text to align we need a somewhat different spacing here than the standard.

PR #499, opened 2020-10-28
Revised modal sizing

teenwolfblitzer:
## Description
Proposed adjustments to the default modal. Current implementation of the modal makes use of `min-width: min-content`. This works as expected for normal content flow, but in the case where a child element is intended to truncate, the `min-content` value prevents the overflow from being hidden (see examples below for WIP examples using the copy text button element).

There doesn't yet appear to be a mockup in Figma, so these changes are likely to be temporary until we have more feedback from design.

Also addresses:
- padding around the modal has been adjusted to match base spacing (`24px`) at mobile viewports
- `button` tags in the template were being closed with anchor link tags
- vertically centers footer elements to match spec in Figma

### Questions:
- should we create "presets" around defined sizes? if so, we'll need to get input from design on what those might be

### Screenshots

| |  before  |  after  |
|---|----|----|
| normal use (padding increased, otherwise no change) |before-sage-modal|after-sage-modal|
| with "copy text button" element on a WIP modal|before-modal-truncate|after-modal-truncate|

### DIFF
--- a/app/views/examples/objects/modal/_markup.html.erb
+++ b/app/views/examples/objects/modal/_markup.html.erb
@@ -5,7 +5,7 @@
       <aside class="sage-modal__header-aside">
         <button class="sage-btn sage-btn--secondary sage-btn--subtle sage-btn--icon-only-dot-menu-horizontal">
           <span class="visually-hidden">Menu</span>
-        </a>
+        </button>
       </aside>
     </header>
     <div class="sage-modal__content">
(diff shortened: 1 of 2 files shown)


app/views/examples/objects/modal/_markup.html.erb
```diff
@@ -5,17 +5,17 @@
       <aside class="sage-modal__header-aside">
         <button class="sage-btn sage-btn--secondary sage-btn--subtle sage-btn--icon-only-dot-menu-horizontal">
           <span class="visually-hidden">Menu</span>
-        </a>
+        </button>
```
philschanely, line 8: 😮 🤦 Nice save!

philschanely (approved):
Makes sense to me. Nice work! 🙌

QuintonJason (approved):
Nice!

PR #504, opened 2020-10-29
Adjust form input message (hint text) spacing

pixelflips:
## Description
Minor adjustment to the input message spacing for the form-input element to match design spec.


teenwolfblitzer (approved):
👍  LGTM!

PR #506, opened 2020-10-29
removing panels from pages

cmcfadzean:
## Description
This PR removes the majority of panels on the different pages
This PR does not touch the Elements & Object pages as I believe we have thoughts about how to better improve these pages in a more broad way.


pixelflips (approved):
Gave each page a quick glance and no concerns here. 👍

PR #510, opened 2020-11-02
Fix busted Dropdown Search attr

AndrwM:
## Description
🥵 I broke the Dropdown with a misnamed attribute for SageSearch.


pixelflips (approved):
👍  LGTM!

PR #512, opened 2020-11-03
Jcm component examples

cmcfadzean:
## Description
Finishing up removing the panels from the Design System Docs and using cards instead. This cleans up the look but makes it simpler to browse. Also moves to use labels in the previews to clean up that look also. I believe that @philschanely is thinking about ways to improve that experience as a whole so didn't refactor everything but gave it as an example.

### Screenshots
Screen Shot 2020-11-02 at 4 54 13 PM

### DIFF
--- a/app/views/examples/elements/_element.html.erb
+++ b/app/views/examples/elements/_element.html.erb
@@ -11,25 +11,25 @@
   do_content: do_content,
   dont_content: dont_content %>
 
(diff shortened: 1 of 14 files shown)


joe-dakroub (approved):
These changes look great; much lighter and easier to process with the use of Cards. The documentation is looking fantastic! 👍

PR #527, opened 2020-11-10
Patch alert links

philschanely:
## Description

Patched alert links so that spacing is only on those inside `__actions`


pixelflips (approved):
LGTM! 👍

PR #529, opened 2020-11-10
Jcm pages

cmcfadzean:
## Description
Now that we have our design Principles this PR creates an area for us to take one more step and showing some of these principles in practice. These are experiences that we believe should be consistent across the app. These guidelines will help us become intentional about how we lean into our design Principles in the actual experience we build.

### DIFF
--- a/app/views/application/_sidebar.html.erb
+++ b/app/views/application/_sidebar.html.erb
@@ -12,6 +12,10 @@ foundations_pages = [
     name: "Accessibility",
     path: pages_foundation_path(:accessibility),
   },
+  {
+    name: "Onboarding",
+    path: pages_foundation_path(:onboarding),
+  },
 ]
 %>
 <%
--- a/app/views/application/_sidebar.html.erb
+++ b/app/views/application/_sidebar.html.erb
@@ -11,11 +11,7 @@ foundations_pages = [
   {
     name: "Accessibility",
     path: pages_foundation_path(:accessibility),
-  },
-  {
-    name: "Onboarding",
-    path: pages_foundation_path(:onboarding),
-  },
+  }
 ]
 %>
 <%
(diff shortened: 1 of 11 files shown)


app/views/pages/crud.html.erb
```diff
@@ -0,0 +1,6 @@
+<%= content_for :heading do %>
+<%= md(%(
+# Create & Distroy
```
pixelflips: Should it be Destroy?

pixelflips (approved):
Noticed one small typo, but overall an awesome addition. 🔥 🔥

app/views/application/_sidebar.html.erb
```diff
@@ -39,6 +39,38 @@ content_pages = [
 ]
 %>
 <%
+experiences_pages = [
+  {
+    name: "Empty States",
```
teenwolfblitzer, line 44: Is there or should there be a particular order here? If not, I'm wondering if this and other menu items could default to alphabetical order, mirroring the elements/objects listings.

app/views/pages/null.html.erb
```diff
@@ -0,0 +1,6 @@
+<%= content_for :heading do %>
+<%= md(%(
+# Empty States
+<p class="docs-heading__lead">How an experience looks without items to display is as important as how it looks with content.</p>
```
teenwolfblitzer, line 4: Surprisingly, "empty states" is the only page with content 🙈

teenwolfblitzer (approved):
👍 LGTM, had one general question about how we order pages in the nav, good to go once the typo @pixelflips mentioned is corrected

app/views/application/_sidebar.html.erb
cmcfadzean: The general idea I had here was tracking the way states would exist in the admin. Empty > then onboarding them > creating > loading during creation.

app/views/pages/null.html.erb
cmcfadzean: I love this so much lol

PR #530, opened 2020-11-11
Jcm icons

cmcfadzean:
## Description
Currently, we have a lot of icons and the list is growing. With that, some icons are being used for various reasons and others are being doubled up. Ideally, this creates guidelines as to how to use icons in the app and what is being used for what.

### DIFF
--- a/app/views/pages/icon.html.erb
+++ b/app/views/pages/icon.html.erb
@@ -66,7 +66,78 @@
 ]
 %>
 <% feature_icons = [
-  "add",
+  {
+    name: "automations",
+    action: "Automations",
+  },
+  {
+    name: "calendar-date",
+    action: "???",
+  },
+  {
+    name: "cart",
+    action: "???",
+  },
+  {
+    name: "chart",
+    action: "Analytics",
+  },
+  {
+    name: "comment",
+    action: "Comments",
+  },
+  {
+    name: "coupon",
+    action: "Coupons",
+  },
+  {
+    name: "drop",
+    action: "Drip",
+  },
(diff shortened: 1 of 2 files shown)


app/views/pages/icon.html.erb
```diff
@@ -148,13 +290,77 @@ text that describes the icon, such as where the icon is purely decorative
...
+      </li>
+      <li class="quick-links__item">
+        <a href="#all-icons" class="quick-links__link">All Icons</a>
+      </li>
+    </ul>
+  </div>
+<% end %>
+<div id="guidelines">
+<%= md(%(
+## General Guidelines
```
philschanely, line 339: FYI We have enabled the Redcarpet table of contents feature that automatically adds an `id` to headings that is a dasherized version of its content. So in this case there is an `id="general-guidelines"` added to this `h2`. Therefore, the `#guidelines` container should not be necessary since the quick links item could point to `#general-guidelines` instead. If indeed you prefer this container, consider adding `class="sage-type"` there and removing the `use_sage_type: true` param on the `md` block below. This way we don't end up with the redundant nested `div`s we have at present 🤓

philschanely (approved):
@cmcfadzean I ❤️ this improvement!! This is super-helpful to clarify and improve on consistent use of icons!! 😍 😍 😍 

One picky comment below before merge for clean code, but otherwise 👍

app/views/pages/icon.html.erb
cmcfadzean: WOW I love this!

PR #539, opened 2020-11-12
Sage Search followup

QuintonJason:
## Description

- [x] - Resolve broken test stemming from adding a `<form>` as the form wrapper instead of a `<div>`
- [x] - updated reset button to be visible on load if the search input have a value

* functionality will be handled on the `kajabi-products` side. Clicking the button does nothing

##### This can only be checked on the app side using the kajabi-products / sage link
1. Visit a page with a form. I use http://www.kajabi.test:3000/admin/products/
2. Enter a value into the search form then press `Enter`
3. When the page refreshes, notice the reset icon is visibe

### DIFF
--- a/lib/sage-frontend/javascript/system/search.js
+++ b/lib/sage-frontend/javascript/system/search.js
@@ -14,21 +14,15 @@ Sage.search = (function() {
   // ==================================================
 
   function init(el) {
-    const elButton = el.querySelector(SELECTOR_SEARCH_BUTTON);
     const elInput = el.querySelector(SELECTOR_SEARCH_INPUT);
-    
+
     // check the input for a value on load
     hasValue(elInput) ? addVisibleButtonState(el) : removeVisibleButtonState(el);
-
-    el.addEventListener('keydown', searchOnInputHandler);
-    elButton.addEventListener('click', searchClearClickHandler);
+    el.addEventListener('keyup', searchOnInputHandler);
   }
 
   function unbind(el) {
-    const elButton = el.querySelector(SELECTOR_SEARCH_BUTTON);
-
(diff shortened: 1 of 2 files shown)


lib/sage-frontend/javascript/system/search.js
```diff
@@ -14,17 +14,15 @@ Sage.search = (function() {
   // ==================================================
 
   function init(el) {
-    const elButton = el.querySelector(SELECTOR_SEARCH_BUTTON);
+    const elInput = el.querySelector(SELECTOR_SEARCH_INPUT);
 
-    el.addEventListener('keydown', searchOnInputHandler);
-    elButton.addEventListener('click', searchClearClickHandler);
+    // check the input for a value on load
+    hasValue(elInput) ? addVisibleButtonState(el) : removeVisibleButtonState(el);
```
AndrwM, line 20: nice!

AndrwM (approved):
+1!

lib/sage-frontend/javascript/system/search.js
```diff
@@ -14,17 +14,15 @@ Sage.search = (function() {
...
+    el.addEventListener('keyup', searchOnInputHandler);
   }
 
   function unbind(el) {
-    const elButton = el.querySelector(SELECTOR_SEARCH_BUTTON);
-
-    el.removeEventListener('keydown', searchOnInputHandler);
-    elButton.removeEventListener('click', searchClearClickHandler);
+    el.removeEventListener('keyup', searchOnInputHandler);
```
AndrwM, line 25: I'm not sure I understand this naming convention. Would `inputHandler` or `keyHanlder` make more sense?

AndrwM (approved):
+1 LGTM!

